PR #24, opened 2026-05-18
feat(shell): enhance Windows shell resolution and fallback logic

EduardtAlbergII:
## Summary
- On Windows, opening a terminal panel failed immediately with `Failed to create terminal: Error invoking remote method 'terminal:create': Error No usable shell found on this system`.
- Root cause: `resolveShell()` in `src/main/shellResolver.ts` had `PLATFORM_FALLBACKS.win32 = []` and an allowlist containing only POSIX shell basenames (`bash`, `zsh`, `sh`, …). With the default `defaultShellPath: ''`, the resolver had nothing to validate, nothing to fall back to, and threw before any PTY was spawned. The inline comment claimed Windows was "not validated by this resolver", but in practice the throw was unreachable to skip.
- The resolver now understands Windows: it allowlists `cmd.exe` / `powershell.exe` / `pwsh.exe` / `bash.exe` / `wsl.exe` (case-insensitive), reads `$COMSPEC` instead of `$SHELL` (which on Windows is typically set by Git Bash / MSYS to a POSIX path node-pty can't spawn), and walks a fallback chain of `$COMSPEC` → PowerShell 7 → Windows PowerShell 5.1 → `cmd.exe`.
- `src/main/shellEnv.ts` already short-circuits on Windows, so no changes were needed there.

### DIFF
--- a/src/main/shellResolver.test.ts
+++ b/src/main/shellResolver.test.ts
@@ -53,15 +53,25 @@ describe('isExecutable', () => {
 describe('resolveShell', () => {
   const originalPlatform = process.platform
   const originalShell = process.env.SHELL
+  const originalComspec = process.env.COMSPEC
+  const originalSystemRoot = process.env.SystemRoot
+  const originalProgramFiles = process.env.ProgramFiles
 
   function setPlatform(p: NodeJS.Platform) {
     Object.defineProperty(process, 'platform', { value: p, configurable: true })
   }
 
+  function restoreEnv(name: string, value: string | undefined) {
+    if (value === undefined) delete process.env[name]
+    else process.env[name] = value
+  }
+
   afterEach(() => {
     Object.defineProperty(process, 'platform', { value: originalPlatform, configurable: true })
-    if (originalShell === undefined) delete process.env.SHELL
-    else process.env.SHELL = originalShell
+    restoreEnv('SHELL', originalShell)
+    restoreEnv('COMSPEC', originalComspec)
+    restoreEnv('SystemRoot', originalSystemRoot)
+    restoreEnv('ProgramFiles', originalProgramFiles)
   })
 
   test('uses the requested path when it is executable', () => {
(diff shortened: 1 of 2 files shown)


PaulHorn (changes requested):
Thanks for the fix — this addresses a real "Cate is unusable on Windows" bug and the fallback chain (`$COMSPEC` → pwsh 7 → PowerShell 5.1 → cmd.exe) is the right call. One blocker before merge:

**The 6 new Windows tests fail when run on macOS/Linux.** I checked this out locally and ran `npm test` on darwin — `Tests 6 failed | 12 passed`, all six failures are in the `Windows` describe block with `No usable shell found on this system`.

Root cause: mocking `process.platform = 'win32'` doesn't change Node's `path` module — on a non-Windows host it keeps POSIX semantics, so `path.basename('C:\\Windows\\System32\\cmd.exe')` returns the entire string (no backslash is treated as a separator), the allowlist check fails, and the resolver throws. Same problem with `path.join` in `windowsFallbacks()` — it produces forward-slash paths on a Mac host, which then don't match the keys the test stubs out.

The runtime is fine on an actual Windows box, but CI / dev machines on Mac/Linux will be red.

Suggested fix — use the explicit platform variants inside the Windows branches:

```ts
// isAllowedBasename
const base = isWindows() ? path.win32.basename(candidate) : path.posix.basename(candidate)

// windowsFallbacks()
fallbacks.push(path.win32.join(programFiles, 'PowerShell', '7', 'pwsh.exe'))
fallbacks.push(path.win32.join(systemRoot, 'System32', 'WindowsPowerShell', 'v1.0', 'powershell.exe'))
fallbacks.push(path.win32.join(systemRoot, 'System32', 'cmd.exe'))
```

(and the same `path.win32.basename` in `rejectionReason` / `pickFallback` for the Windows path).

That keeps the runtime identical on Windows and makes the new tests deterministic on any host, so the regression you're guarding against doesn't silently come back later. Happy to merge once `npm test` is green.

PaulHorn (approved):
LGTM. Nice root-cause writeup — the /bin/zsh→ swap is the key insight (Git Bash leaking /usr/bin/bash into /bin/zsh on Windows would have bitten anyone trying to add a Windows fallback chain without that swap). Allowlist + case-insensitive basename match + explicit path.win32/path.posix usage keeps the cross-platform test stubbing honest. Good coverage of the empty-settings regression case.

PR #93, opened 2026-05-25
fix: resolve Windows agent launch crash (spawn node ENOENT)

### LINKED ISSUE #92: Windows: agent launch crashes with 'spawn node ENOENT' (node.cmd shim not resolvable by shell-less spawn)
## Description

On Windows, launching the agent (e.g. clicking to create/buy an agent / pick a model) crashes the app with a native Electron dialog:

```
A JavaScript error occurred in the browser process
Error: spawn node ENOENT
    at ChildProcess._handle.onexit (node:internal/child_process:287:19)
    at onErrorNT (node:internal/child_process:508:16)
    at process.processTicksAndRejections (node:internal/process/task_queues:90:21)
```

## Root cause

The agent is launched by the third-party `@earendil-works/pi-coding-agent` RPC client, which hardcodes:

```js
this.process = spawn("node", [cliPath, ...args], { /* shell: false (default) */ })
```

(`node_modules/@earendil-works/pi-coding-agent/dist/modes/rpc/rpc-client.js:40`)

To make `node` resolvable in a packaged app, `src/agent/main/agentManager.ts` (`buildAgentEnv`) creates a shim directory and prepends it to `PATH`. On Windows that shim is a **`node.cmd`** batch wrapper (`src/agent/main/nodeShim.ts`).

The problem: a shell-less `spawn("node", ...)` on Windows resolves only `node` / `node.exe` via `CreateProcess`; it does **not** resolve `.cmd` files (those require a shell). So `node` is never found on `PATH`, `spawn` emits `ENOENT`, the error is uncaught in the main process, and Electron shows the crash dialog.

The `node.cmd` shim therefore never works for this caller on Windows.

## Proposed fix

Make `node` resolvable without a shell by creating a real **`node.exe`** in the shim directory that points at the Electron binary (`process.execPath`), keeping `ELECTRON_RUN_AS_NODE=1` in the spawned env:

- Prefer a hardlink (`fs.linkSync`) to `process.execPath` (cheap, no admin/Developer Mode needed when on the same volume).
- Fall back to a copy (`fs.copyFileSync`) if the hardlink fails (e.g. shim temp dir is on a different volume than the app).

## Environment

- OS: Windows 11
- Installed from the latest GitHub release.

NicolaasGrobler:
## Summary

Fixes #92.

On Windows, launching the agent crashed the app with a native Electron dialog: `Error: spawn node ENOENT`.

pi's `RpcClient` launches the agent via a shell-less `spawn("node", [cliPath, ...])`. On Windows, `CreateProcess` only resolves `node` / `node.exe` on `PATH` — it never resolves a `.cmd` file (those require a shell). The previous shim wrote a `node.cmd` wrapper, which was therefore invisible to that spawn, producing `spawn node ENOENT` and crashing the main process.

## Fix

`src/agent/main/nodeShim.ts`: on Windows, create a real `node.exe` in the shim directory instead of `node.cmd`:

- Hardlink `node.exe` to the Electron binary (`process.execPath`) — cheap and needs no Developer Mode/admin when the shim dir and binary are on the same volume.
- Fall back to a copy (`fs.copyFileSync`) when a hardlink isn't possible (e.g. different volumes).

`ELECTRON_RUN_AS_NODE=1` is already set in the spawned env by `buildAgentEnv()`, so the Electron binary behaves as Node. macOS/Linux behavior (symlink) is unchanged.

## Test plan

- [x] `src/agent/main/nodeShim.test.ts` updated and passing, including a Windows integration test that shell-less-spawns the `node.exe` shim (`execFileSync`) and asserts it runs as Node (prints `ok`).
- [x] `npm run build` passes.
- [ ] Manual: on a packaged Windows build, launch an agent / pick a model — no `spawn node ENOENT`, agent starts.

## Notes

- Best-effort temp-dir cleanup was added in the test because a `node.exe` hardlinked to the running binary can be briefly locked by Windows (EPERM) after the spawned child exits; leaving the temp dir is harmless.

### DIFF
--- a/src/agent/main/nodeShim.test.ts
+++ b/src/agent/main/nodeShim.test.ts
@@ -9,7 +9,18 @@ function makeTmpDir(): string {
 }
 
 function cleanup(dir: string) {
-  fs.rmSync(dir, { recursive: true, force: true })
+  // Best-effort: a node.exe hardlinked to the running binary can stay briefly
+  // locked on Windows (EPERM). Leaving the temp dir behind is harmless.
+  try {
+    fs.rmSync(dir, { recursive: true, force: true })
+  } catch { /* temp dir — let the OS reclaim it */ }
+}
+
+/** Write a fake executable file (the shim's link/copy target) and return its path. */
+function makeFakeExe(dir: string, name: string, contents: string): string {
+  const exe = path.join(dir, name)
+  fs.writeFileSync(exe, contents)
+  return exe
 }
 
 describe('createNodeShim', () => {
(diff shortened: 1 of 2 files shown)


Anton-Horn:
Nice fix — the .cmd → .exe swap makes total sense. CreateProcess ignoring .cmd for shell-less spawns is one of those Windows gotchas you don't see coming until it bites you.

Checked the CI runs, all three platforms green including the windows-latest runner where the shell-less spawn test actually exercises the real codepath. Hardlink-with-copy-fallback is a clean approach too.

LGTM, merging.

Anton-Horn:
Thanks, looks good! Added one more integration test for this.

PR #127, opened 2026-05-27
feat(palette): unified accent palette for workspaces + canvas regions

schroejahr2:
## Summary

Workspaces and canvas regions used to live on two independent color palettes — six muted hex accents for the workspace context menu and eight pure-rgba fills for the region picker. The lists drifted in count, order, and hue choices, with no path to keep them in sync.

This PR introduces a single source of truth in `src/shared/colors.ts`:

- `ACCENT_PALETTE` — eight slots, each pairing a **muted hex** (workspace tab accent) with a **saturated rgb tuple** (region fill).
- Shared slot order — Blue first, then a rainbow run through Red, Orange, Yellow, Green, Teal, Purple, Pink. Same logical hue at the same index in either picker.
- Both pickers now read names from `ACCENT_COLOR_NAMES`, so renaming a slot updates every menu at once.

### Wiring

- `appStore.ts` re-exports `WORKSPACE_COLORS` from the shared palette.
- `WorkspaceTab.tsx` drops its local `COLOR_NAMES` and uses `ACCENT_COLOR_NAMES`.
- `CanvasRegionComponent.tsx` drops its local `REGION_COLORS`/`REGION_COLOR_LABELS` and uses `REGION_FILL_COLORS` + `ACCENT_COLOR_NAMES`. Resize-handle backgrounds now inherit the region's color instead of a hardcoded blue.
- `canvasStore.addRegion` defaults to `REGION_FILL_COLORS[0]`.

### Workspace accent in the dock tab strip

`MainWindowShell.tsx` publishes the active workspace's color as a `--workspace-accent` CSS variable on the shell root. The dock tab-bar's active-tab top stripe (`DockTabBar.tsx:193`) now resolves:

```
var(--workspace-accent, var(--node-chrome-accent, #3b82f6))
```

Workspace color wins when set, the per-terminal `--node-chrome-accent` is the fallback when no workspace color is configured, and #3b82f6 is the final fallback. The cascade was a deliberate choice — terminal-theme accents stay available for un-colored workspaces, and a colored workspace gets a consistent stripe across both the canvas dock and every canvas-node mini-dock inside it. Detached `PanelWindow`/`DockWindow` shells don't set `--workspace-accent`, so they fall back to the existing behaviour.

### Glow — considered, deferred

We discussed extending the workspace tint to the focus glow on canvas nodes (currently a hardcoded `rgba(74, 158, 255, ...)` halo in `useCanvasNodeStyle.ts:15`). Decided to leave it alone for now — the stripe already telegraphs the workspace identity, the glow's very-low alpha makes it cosmetically neutral, and recoloring it risks making heavily-populated canvases feel washed in workspace color. Easy to revisit if we want it.

### Backward compatibility

Legacy persisted `region.color` strings keep working — `parseRgba` and the minimap regex both handle any well-formed `rgba(...)` value. The "Change Color" context menu just shows no ✓ for old values until the user picks a new one.

## Test plan

- [x] Right-click a workspace → menu shows 8 colors with shared names.
- [x] Right-click a canvas region → same 8 names, same order.
- [x] Pick a workspace color → active dock tab + every canvas-node mini-dock tab show the workspace color on the top stripe.
- [x] Pick a region color → fill + resize handles + selection border all tint to the picked hue.
- [x] Type-check + drag/docking tests unaffected (`npm test`).

### DIFF
--- a/src/renderer/canvas/CanvasRegionComponent.tsx
+++ b/src/renderer/canvas/CanvasRegionComponent.tsx
@@ -4,18 +4,7 @@ import { useCanvasStoreContext, useCanvasStoreApi } from '../stores/CanvasStoreC
 import type { NativeContextMenuItem } from '../../shared/electron-api'
 import { useAppStore, getCanvasOpsById } from '../stores/appStore'
 import { confirmDeleteRegion } from '../lib/confirmDeleteRegion'
-
-// Preset region colors
-const REGION_COLORS = [
-  'rgba(0, 128, 255, 0.08)',    // blue (default)
-  'rgba(0, 255, 0, 0.08)',      // green
-  'rgba(255, 128, 0, 0.08)',    // orange
-  'rgba(255, 0, 0, 0.08)',      // red
-  'rgba(170, 0, 255, 0.08)',    // purple
-  'rgba(255, 255, 0, 0.08)',    // yellow
-  'rgba(0, 255, 255, 0.08)',    // teal
-  'rgba(255, 0, 128, 0.08)',    // pink
-]
+import { ACCENT_COLORS, ACCENT_COLOR_NAMES, REGION_FILL_COLORS } from '../../shared/colors'
 
 // Parse rgba(...) string → { r, g, b, a }
 function parseRgba(str: string): { r: number; g: number; b: number; a: number } {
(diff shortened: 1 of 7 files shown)


Anton-Horn:
Small follow-up: the `parseRgba` fallback in `CanvasRegionComponent.tsx` still uses the old hardcoded `{ r: 74, g: 158, b: 255, a: 0.08 }` — would be nice to pull that from `REGION_FILL_COLORS[0]` / `REGION_FILL_ALPHA` so the fallback stays in sync with the palette.

Also, the remaining hardcoded `rgba(74, 158, 255, ...)` values in drag/selection UI (`ZoneIndicator`, `Overlay`, `SnapGuides`, `Canvas` marquee, `useCanvasNodeStyle` focus glow) are out of scope here but would be a natural follow-up — #129 could pick those up when the theme accent becomes configurable.

Issue #122, opened 2026-05-27
Add support for apps natively


bhadraagada:
It's really good concept but one thing whuihc can make this application the next thing could support for application where i can use my current stack ie.:
- warp for terminal
- cursor for ide and agent
- opencode for ai agent
- zen for browser
- helium for browser

so i will be able to open these custom windows and not stuck to limited set of options provided by u guys which is limiting initial testing and exploring of this application.

if u want we can connect for this feature and work this thing together and make this happen

PaulHorn:
Thanks for the suggestion! This is an interesting direction.

Embedding native app windows into the canvas is technically feasible — the approach would be to launch the app as a child process, grab its window handle, and reparent it into Electron's window (`SetParent` on Windows, `NSWindow addChildWindow` on macOS, `XReparentWindow` on Linux/X11).

We already have some of the building blocks in place (native addons, the panel/canvas system, process management via node-pty), so a `NativeAppPanel` type that wraps reparented windows isn't out of reach.

The main challenges:
- **Window handle discovery** — reliably mapping a launched PID to its window handle varies by platform
- **Position sync** — the embedded window needs to follow canvas pan/zoom/resize without flicker
- **Focus management** — native windows and Electron compete for focus
- **Platform coverage** — each OS needs its own native code path (Wayland on Linux is especially tricky)

We're putting together a feature timeline over the next few days, and this will definitely be on it to explore feasibility. We'll reach out!

bhadraagada:
i feel like we have an headless wrapper or something for this where we can just drop in the windows tht we want to include in the infinite space.

ig for custom windows this will the best approach this way wont have to deal with wired interactions tht u mentioned earlier

@PaulHorn

galori:
+1 to this concept - this is what I was hoping Cate would do when I installed it. 

I can't even log in to my work's github through the built in browser because it requires an Okta extension and network communications just to auth, and I would want a browser where I can open dev-tools when I need to.

robinwettstaedt:
I also thought this was the main feature of Cate, essentially an endless canvas for all apps. Exciting concept, will check back every once in a while to see if you added that!

PaulHorn:
@galori @robinwettstaedt we noticed that too, and it’s definitely something we want to focus on more.
A better browser experience, devtools, auth flows like Okta, and eventually bringing more apps onto the canvas are all very relevant for where Cate should go.
[...]

PR #131, opened 2026-05-27
feat: add native Windows app panels

bhadraagada:
Add a Windows-only native app panel that can launch or attach external apps and keep their windows synced over the canvas, so users can bring their existing tools into Cate.

## note
this is a very basic thing which i managed to make in like one pass 
this can be refined further.

#122

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,4 @@ test-results/
 
 # Claude Code local settings
 .claude/settings.local.json
+.tmp-pkgs/
--- a/package-lock.json
+++ b/package-lock.json
@@ -23,6 +23,7 @@
         "electron-log": "^5.4.3",
         "electron-store": "^10.0.0",
         "electron-updater": "^6.8.3",
+        "koffi": "^3.0.2",
         "mammoth": "^1.12.0",
         "monaco-editor": "^0.52.0",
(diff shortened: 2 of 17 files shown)


schroejahr2:
in consideration #126

Anton-Horn:
We want to have same behaviour on all platforms. For Windows this feature is somewhat straight forward to implement. Any idea's how you would continue for Mac / Linux?

bhadraagada:
Yeah, I’m thinking less “true embed/reparent” and more a lightweight headless wrapper / app-host model.

The idea #122 is: Cate owns the panel + canvas behavior, while the external app stays a normal OS window. The host layer just launches/attaches to that window and keeps its screen-space bounds synced with the Cate panel.

That should avoid a lot of the weird focus/input/reparenting issues with arbitrary apps like Cursor, Warp, Zen, etc. The shared abstraction could be something like a `NativeWindowProvider` / `AppHost`, with platform-specific implementations underneath:

- Windows: HWND + `SetWindowPos`
- macOS: Accessibility API / `AXUIElement` move+resize
- Linux: X11 via Xlib/wmctrl first
- Wayland: likely limited because compositors restrict global window control

So the behavior can be consistent from Cate’s side, while each OS handles the actual window control differently.

PR #144, opened 2026-05-27
fix(agent): drive running state from PTY streaming + post-startup subprocesses (#140)

### LINKED ISSUE #140: Agent running detection in terminal is inconsistent
The detection of whether an agent is running in a terminal is inconsistent — it only shows as "running" when a tool call is active, not throughout the entire agent session.

## How it works today

The main process polls the process tree every 1s (`src/main/ipc/shell.ts`). It matches known agent process names (claude, codex, gemini, etc.) and sets `agentPresent`. Then `agentIsActive()` walks the agent's child processes — if it finds subprocesses (npm, git, python, etc.), it reports `subprocessActive = true`.

The renderer (`src/renderer/lib/agentScreenDetectorLogic.ts`) maps this to state:
- **running** → agent present AND subprocess active
- **waitingForInput** → agent present, no active subprocesses (3s hysteresis)
- **notRunning** → agent process not found

## The problem

When Claude is thinking or streaming a response (no child processes spawned), `subprocessActive` is false → state drops to `waitingForInput`. The "running" indicator only lights up during tool execution, making it feel flickery and unreliable.

## Expected behavior

The agent should show as "running" throughout the entire session — from launch to exit — with sub-states for "executing" vs "thinking" if needed.

schroejahr2:
Fixes #140.

## Problem

Sidebar "agent running" shimmer was unreliable:

- **Original bug:** dropped to `waitingForInput` mid-stream because the detector only tracked `subprocessActive` (presence of a child PID), and during pure model output / thinking there is *no* subprocess — just bytes flowing through the PTY.
- **MCP regression after the first fix:** users running `claude` with MCP servers (e.g. `context7-mcp`, `bun`-based MCP runtimes) had `subprocessActive=true` forever, so the shimmer never stopped even when the agent was idle waiting for input.

## Fix

Two independent activity signals, OR'd in `computeRawState`:

1. **`isStreaming`** — PTY bytes/s above a threshold (`STREAMING_BYTES_PER_INTERVAL = 200`).
   Catches model output and the thinking spinner without depending on a subprocess. Uses a cumulative byte counter (`totalPtyBytes`) on `TerminalIdleState` so resume-from-idle does *not* reset the baseline.
2. **`subprocessActive`** (age-filtered) — true only for children spawned **after** a startup grace window.
   MCP servers + their `bun`/`node` runtimes appear within the first few seconds after the agent process and are filtered out; tools spawned later (silent or noisy) count.

State machine unchanged otherwise; 3 s hysteresis still holds `running → waitingForInput` to avoid flicker.

## Agent detection

Detects `claude` / `claude-code`, `codex`, `antigravity`, `cursor` / `cursor-agent`, `opencode`, `forge` / `forgecode`, `pi`, and `gemini` (legacy).

> **Note:** Gemini CLI is **deprecated in favor of Antigravity CLI**. See [Google's announcement](https://developers.googleblog.com/an-important-update-transitioning-gemini-cli-to-antigravity-cli/). The `gemini` entry is kept for legacy installs.

## How the helper filter works

`src/main/ipc/agentPidAge.ts` tracks first-seen timestamps per child PID.

- On **first sight** of a child, `shell.ts` calls `ps -o etime=` to read the child's real start time and uses that as `firstSeen` — important when reattaching to an already-running agent, otherwise startup helpers would look "fresh" and get classified as tools.
- `parseEtime()` handles BSD `etime` format (`MM:SS` / `HH:MM:SS` / `DD-HH:MM:SS`) for macOS portability. Linux `etimes` (seconds) is not portable across the two.
- A child first-seen ≤ `agentStartedAt + STARTUP_GRACE_MS` (5 s) is a helper. Anything later is a tool, regardless of how long it then lives.
- Name-pattern filter (`*-mcp`, `mcp-*`, known idle helpers) catches the obvious cases too.

`bytesSamples` is dropped per terminal when the agent disappears so a restart starts with a clean baseline. `lastSeenPidsByTerminal` carries pids across skipped/errored scan cycles so the prune step doesn't wipe valid entries.

## Tradeoff

This is a heuristic, not a guarantee. Thresholds (200 B/s, 5 s grace window) are empirical and may need tuning for edge cases (slow streams, slow-starting MCPs, bursty long-running tools). We measure *symptoms* (bytes, subprocesses), not agent state.

The only truly reliable solution would be **hooks** — an explicit protocol where each agent CLI emits lifecycle events (`thinking`, `awaiting_input`, `tool_invoked`, `done`). That doesn't exist across agents today. If it ever does, it should replace this heuristic entirely.

Good enough for daily use right now; not a final answer.

## Files of interest

- `src/main/ipc/shell.ts` — main detection logic, `agentIsActive(terminalId)`, `childFirstSeenAt`, `isPostStartupChild`, `parseEtime`, `AGENT_DEFINITIONS`.
- `src/main/ipc/agentPidAge.ts` (new) — PID first-seen map + prune.
- `src/main/ipc/terminal.ts` — `totalPtyBytes` cumulative counter, `getTotalPtyBytes()`.
- `src/renderer/lib/agentScreenDetectorLogic.ts` — `DetectorSignals` with `isStreaming` + `subprocessActive`.
- `src/renderer/lib/agentScreenDetector.ts` — drops `finished`-state notification.
- `src/renderer/stores/statusStore.ts`, `src/renderer/hooks/useProcessMonitor.ts`, `src/preload/index.ts`, `src/shared/electron-api.d.ts` — plumb `isStreaming` through IPC.

## Test plan

- [x] `npm test` — 328 pass (2 pre-existing git-env failures, environmental)
- [x] Manual dev: `claude` with MCPs → shimmer on while streaming/thinking → stops within hysteresis window after response → dot + "waiting for input" notification → no notification on agent exit
- [x] Manual dev: `codex` — same behavior
- [x] Reattach to long-running agent (close+reopen window) — verify helpers still classified as helpers via `etime` backdating

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,6 @@ test-results/
 
 # Claude Code local settings
 .claude/settings.local.json
+
+# Local logo scratch dir
+_logos/
--- a/src/main/ipc/agentPidAge.test.ts
+++ b/src/main/ipc/agentPidAge.test.ts
@@ -1,0 +1,42 @@
+import { describe, expect, it, beforeEach } from 'vitest'
+import { STARTUP_GRACE_MS, __resetPidAges, recordPidFirstSeen, pruneStalePidAges } from './agentPidAge'
+
+describe('agent pid first-seen tracking', () => {
+  beforeEach(() => { __resetPidAges() })
+
+  it('records first sight and returns that timestamp', () => {
+    expect(recordPidFirstSeen(123, 1_000)).toBe(1_000)
+  })
+
+  it('returns the original timestamp on later calls (caller passes actual start time)', () => {
+    recordPidFirstSeen(123, 1_000)
+    expect(recordPidFirstSeen(123, 99_999)).toBe(1_000)
+  })
+
+  it('respects an explicit backdated first-seen for already-running PIDs', () => {
+    // simulates shell.ts passing the actual process start time on first sight
+    expect(recordPidFirstSeen(7, -50_000)).toBe(-50_000)
+  })
+
+  it('pruneStalePidAges drops pids missing from the cycle set', () => {
+    recordPidFirstSeen(1, 1_000)
+    recordPidFirstSeen(2, 1_000)
+    pruneStalePidAges(new Set([1]))
+    // pid 2 was pruned → re-records against a new timestamp
+    expect(recordPidFirstSeen(2, 5_000)).toBe(5_000)
+    // pid 1 was kept → original timestamp preserved
+    expect(recordPidFirstSeen(1, 9_999)).toBe(1_000)
+  })
+
+  it('startup grace covers helpers spawned right after the agent', () => {
+    const agentStartedAt = 0
+    const helperFirstSeen = recordPidFirstSeen(10, STARTUP_GRACE_MS - 1)
+    expect(helperFirstSeen - agentStartedAt).toBeLessThanOrEqual(STARTUP_GRACE_MS)
+  })
+
+  it('post-grace children fall outside the helper window', () => {
+    const agentStartedAt = 0
+    const toolFirstSeen = recordPidFirstSeen(20, STARTUP_GRACE_MS + 1)
+    expect(toolFirstSeen - agentStartedAt).toBeGreaterThan(STARTUP_GRACE_MS)
+  })
+})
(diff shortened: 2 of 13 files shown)


Anton-Horn:
Noticed the tab flips to running while I'm typing into Claude Code — the input box repaints on every keystroke and that goes over the 200 B/s threshold. Stop typing and it drops back to awaiting input after the 3s hold. The byte-rate signal can't really tell thinking apart from repainting my keystrokes. Not a blocker, just flagging the edge.

schroejahr2:
> Noticed the tab flips to running while I'm typing into Claude Code — the input box repaints on every keystroke and that goes over the 200 B/s threshold. Stop typing and it drops back to awaiting input after the 3s hold. The byte-rate signal can't really tell thinking apart from repainting my keystrokes. Not a blocker, just flagging the edge.

please see my comment in #140 currently its the best bet on my system, milage can vary

PR #148, opened 2026-05-28
feat(panels): agent-aware terminal tab UI (icon, title, awaiting-input dot)

schroejahr2:
> Follow-up to #144. Builds on the per-terminal agent detection that PR shipped.

## Summary

Three independent additions to make a terminal that's running an agent CLI feel like its own first-class panel, instead of a generic "Terminal N".

### a) Tab icon = agent logo

When the per-terminal detector finds an agent CLI in the process tree, the dock tab and the sidebar workspace tree swap the generic Phosphor `Terminal` glyph for the agent's logo.

Supported (matched against `displayName` from `AGENT_DEFINITIONS` in `src/main/ipc/shell.ts`):

- Claude Code · Codex · Antigravity · Cursor · OpenCode · Forge Code · Gemini CLI · PI Agent

Falls back to the Phosphor Terminal glyph when no agent is detected or the logo asset fails to load (``). Logo lookup is gated on `agentPresent` so the icon flips back the moment the agent process exits — `shell.ts` deliberately keeps `agentName` populated after exit so the status footer can still say "Finished (Claude Code)".

Logos live under `src/renderer/assets/agentLogos/` (moved from the gitignored `_logos/` so Vite bundles them). Sources: **Wikimedia Commons** and **dblogo.com**.

### b) Terminal title from the agent

`terminalRegistry` now subscribes to `xterm.onTitleChange` on both the fresh-spawn and reconnect paths. `agentTitleParser` extracts the middle em-dash segment of OSC 0/1/2 titles so the tab shows the live status the agent emits, e.g.:

```
georgschrojahr — ✱ Test schroejahr.de aufrufen — bun ‹ claude — 133×24
                └──────────── used as the tab title ──────────┘
```

Pushed to a new appStore action `updatePanelTitleFromAgent`. A new `titleUserOverridden: boolean` field on `PanelState` blocks OSC updates once the user manually renames the tab via `renamePanelByUser`.

Codex doesn't emit OSC titles, so `useProcessMonitor` also pushes the agent display name as a starter title on the rising edge (null → "Codex"). Agents that do emit OSC overwrite this immediately with their live status string.

OSC updates are deferred via `queueMicrotask` so OSC sequences flushed from the scrollback during `terminal.open()` don't run `set()` inside React's commit phase (would otherwise trip "Maximum update depth").

### c) Awaiting-input dot in dock tab

Existing `cate-await-indicator` CSS (already used by the sidebar) now also renders in the dock tab, left of the close button, when `agentState === 'waitingForInput'` for that panel. Matches the sidebar's tan accent (`#c08a5a`).

## Notes

- `lastAgentName` map in `useProcessMonitor` is cleared by `statusStore.unregisterTerminal` so it can't grow without bound across long dev sessions.
- Detached panel/dock windows (`PanelWindowShell`, `DockWindowShell`) don't subscribe to `useProcessMonitor`, so the agent state and logo don't sync there yet. Known limitation, separate iteration.

## Image attribution

- Logo SVG/PNG assets sourced from **Wikimedia Commons** and **dblogo.com**.

## Test plan

- [x] `npm test` — 335 pass (2 pre-existing git-env failures)
- [x] Manual dev: claude / codex / gemini → logo swaps, title tracks OSC, dot appears on waiting
- [x] Agent exit → logo falls back to Phosphor Terminal glyph
- [x] Switch agents in same shell (claude → codex) → logo updates
- [x] User-rename a tab → subsequent OSC updates are ignored

Bildschirmfoto 2026-05-28 um 11 40 45

Bildschirmfoto 2026-05-28 um 10 40 15


Anton-Horn:
Pushed a small refactor: the dock tabs and sidebar were each re-deriving the ptyId→panelId mapping + the agentPresent gate by hand (three copies). Pulled it into one useAgentInfoByPanel hook, scoped per workspace. Not fewer lines overall, but the invariant lives in one place now instead of drifting across call sites.

PR #151, opened 2026-05-28
feat(browser): widen resize hit area and add device size presets

### LINKED ISSUE #147: Browser resize (manual and automatic)
Hey !

Love the Cate project, I think it's super cool !!

One thing that could definitely worth improving imo is the resizing of browser.
First the area when cursor switch to click to resize seems to be super small, plus that would be awesome to have like custom presets of size (by system or custom, both would work for me), like mobile, big screen, ipad, ...
This could help a lot!

I haven't seen any way to do so.

Thanks for this project!

AlexisBnnft:
Closes #147.

## What

Two small UX improvements to the browser panel resize, requested in issue #147.

### 1. Wider resize hit area (6 → 10 px)

The `RESIZE_THRESHOLD` constant in `useNodeResize.ts` is bumped from 6 to 10 pixels. The 6px band was uncomfortably small in general and especially around the browser panel: the `<webview>` captures pointer events, shrinking the usable edge to whatever sits between the node's outer border and the webview's edge.

### 2. Edge guards over the webview

Transparent overlay divs are placed on top of the webview's borders and corners. They forward `mousedown` to the CanvasNode resize handler (events would otherwise be captured by the webview) and set the correct resize cursor for that side/corner. Together with the wider threshold, the browser panel now feels as easy to resize from any edge as the other panels.

### 3. Device size presets

A new "Devices" button is added next to the Screenshot button in the URL bar. It opens a small menu with these presets:

| Label    | Size       | Hint              |
| -------- | ---------- | ----------------- |
| Mobile   | 375×667    | iPhone SE         |
| Mobile L | 414×896    | iPhone 11 Pro Max |
| Tablet   | 768×1024   | iPad              |
| Tablet L | 1024×1366  | iPad Pro          |
| Laptop   | 1280×800   |                   |
| Desktop  | 1920×1080  |                   |

Selecting a preset calls `canvasStore.resizeNode` and accounts for the URL bar height so the visible viewport matches the preset exactly. The active preset is marked with a check.

## Why

Both pieces of feedback come from #147 — the small hit area was frustrating in normal use and made browser resize harder than for other panels, and there was no quick way to test responsive layouts at common viewport sizes without manually dragging to approximate dimensions.

## Test

- `npm run build` passes locally.
- Verified manually with `npm run dev` on macOS (arm64): resize feels comfortable on all four edges and all four corners; preset menu opens/closes correctly; selecting a preset resizes the node exactly to the requested viewport (URL bar accounted for); the active preset shows a check mark; clicking outside closes the menu.

## Notes

- No new dependencies. `Devices` and `Check` icons are already in `@phosphor-icons/react` (already used elsewhere in BrowserPanel).
- The threshold change is global (affects all panels), which I think is a net improvement but happy to scope it to the browser if maintainers prefer.
- Screenshots can be added if useful — let me know.

### DIFF
--- a/src/renderer/hooks/useNodeResize.ts
+++ b/src/renderer/hooks/useNodeResize.ts
@@ -58,7 +58,11 @@ interface UseNodeResizeReturn {
 // Edge detection (exported for use by CanvasNode)
 // -----------------------------------------------------------------------------
 
-const RESIZE_THRESHOLD = 6
+// Edge / corner detection band. Wider than the original 6px to give a more
+// comfortable hit target — especially important for the browser panel where the
+// <webview> captures pointer events and the usable resize strip is the band
+// between the node's outer edge and the webview's edge. See issue #147.
+const RESIZE_THRESHOLD = 10
 
 /**
  * Detect if a mouse position (relative to the node's top-left) is near an
--- a/src/renderer/panels/BrowserPanel.tsx
+++ b/src/renderer/panels/BrowserPanel.tsx
@@ -5,14 +5,25 @@
 // =============================================================================
 
 import { useEffect, useRef, useState, useCallback } from 'react'
-import { Globe, ArrowLeft, ArrowRight, ArrowClockwise, Camera, MagnifyingGlass } from '@phosphor-icons/react'
+import { Globe, ArrowLeft, ArrowRight, ArrowClockwise, Camera, MagnifyingGlass, Devices, Check } from '@phosphor-icons/react'
 import { useSettingsStore } from '../stores/settingsStore'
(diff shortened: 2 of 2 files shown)


Anton-Horn:
Decided to go with #149 for the resize fix, it's generic across all panels and has the corner/edge thresholds plus tests. The device-size presets here are a nice addition though; worth splitting into their own PR. One thought for that: maybe aspect ratios rather than hardcoded px values, so presets scale with the panel instead of snapping to fixed dimensions.

AlexisBnnft:
Will redo !

PR #153, opened 2026-05-28
refactor(agent): detect running state from the agent's spinner

Anton-Horn:
Follow-up on #148 (stacked: #144 → #148 → this). Replaces #144's byte-rate + post-startup-subprocess heuristic with a signal the agent itself emits — its animated spinner.

## Why
#144's detector inferred "running" from PTY byte volume + age-filtered child processes. During model *thinking* there's little output and no subprocess, so it dropped to `waitingForInput` mid-turn and fired false "needs input" notifications. The spinner animates *throughout* a turn (including thinking), so it's a true busy signal.

## How
"running" = a spinner glyph is animating; idle = its absence past a short settle window.
- **claude / codex** — braille glyph (U+2800–U+28FF) in the OSC title.
- **pi / cursor / agy** — braille glyph in the screen body.
- **opencode** — block scanner bar; its inactive cell `⬝` (U+2B1D) is the unique, idle-safe tell.

Each was validated against the agent's real captured terminal output (fires while working, absent at steady idle).

## Also in here
- **Consistent agent tab titles.** Agent terminals show the detected agent name (`Codex`, `Claude Code`, …) instead of each agent's inconsistent raw OSC title (codex → cwd, claude → `✳ Claude Code`, others → session labels). Plain shells still show their OSC title (cwd).
- **Antigravity match fixed** to its real `agy` binary — `antigravity` was dead (that's the GUI IDE's Electron process, never a terminal child).
- **Removed Gemini CLI and Forge** from supported agents.

## Removed (net −414 lines)
`agentPidAge`, the PTY byte counter, the 200 B/s streaming sampler, the 250 ms polling tick, the hysteresis hold, and `notificationTransitions` — plus the `subprocessActive`/`isStreaming` store fields and IPC args.

## Notes / limitations
- Signal rests on each agent's (undocumented) spinner style — version drift could break a given agent; it fails *safe* (no indicator rather than a wrong one).
- agy/codex animate a spinner during startup/sign-in, so a freshly-attached agent may read "running" briefly.
- Verified at the logic/byte level against real captures; not exercised in the live Electron UI.

## Merge order
Stacked on #148. Merge #144 → #148 → this. (#148 already contains #144, so merging #148 also lands #144.)

## Test plan
- [x] `tsc --noEmit` clean
- [x] `vitest run` — 304 pass (2 pre-existing git-env failures), incl. new spinner + coordinator-timing tests
- [ ] Manual: each agent shimmers while working, switches to the awaiting dot after finishing, tab shows the clean agent name


schroejahr2:
I see where you are coming from. 

The problem I face with this is that it breaks support for long lasting sessions like /monitoring or watchdogs which have no spinner and will always display as "input required"

PR #155, opened 2026-05-28
feat: rationalize file exclusions by making configurable; preserved strong opinionated exclusions

cbergoon:
## Summary
- Adds a File Explorer section in settings to edit the list of folder/file names hidden from the file explorer, file search, and change watching. The list persists across sessions and applies to every project.
- Exclusions are read live in the main process, so edits take effect without a relaunch: open explorers refresh via a new `SETTINGS_CHANGED` broadcast, and active filesystem watchers are rebuilt with the new ignore globs.
- Trims the built-in defaults: `dist` and `build` are no longer hidden by default (common names for hand-written source/config that exact-name matching could
silently hide). Remaining defaults: `.git`, `.DS_Store`, `.Trash`, `node_modules`, `__pycache__`, `.npm`, `.cache`, `.build`, `.swiftpm`, `DerivedData`, `Pods`.
    
## Implementation
- `fileExclusions: string[]` added to `AppSettings`/`DEFAULT_SETTINGS`, persisted via electron-store (schema entry in `store.ts`).
- `filesystem.ts` reads the setting live (`getSettingSync`) in `readDir`, `searchFiles`, and the watcher; new `refreshWatcherIgnores()` rebuilds pooled watchers on change.
- New `SETTINGS_CHANGED` channel + `onSettingsChanged` preload bridge; `FileExplorer` reloads the tree when `fileExclusions` changes.


PaulHorn:
Nice work, making the exclusion list configurable is the right call and matches what every serious editor does (VS Code, JetBrains, Zed, Sublime all ship an editable list). Trimming dist/build from the defaults also makes sense, since those names are sometimes hand-authored rather than generated.

One design thing I wanted to raise before this lands, not a blocker, more something I'd like us to decide consciously. Right now the PR drives three different concerns off a single list: what shows in the explorer, what file search looks at, and what the chokidar watcher tracks. VS Code keeps those as three separate lists on purpose, because they don't always line up. The classic case is node_modules: sometimes you genuinely want to browse a dependency's source in the tree, but the moment you remove it from exclusions to see it, you also start watching all of it again, which is the exact perf cost the exclusion was there to avoid. There's no way to say 'show it, but don't index or watch it.'

A single combined list is a perfectly reasonable place to start and probably fine for now, since most people won't hit that edge. I'm mostly wondering whether the plan is to split visibility from search/watch later if it comes up, or whether the single list is the intended long-term model. Either answer is fine, just worth a sentence here so it's a deliberate choice rather than something we backed into.

PaulHorn (commented):
Did a proper pass over this — overall it's a clean, well-thought-out PR. The `createWatcher` extraction with the subscribers map captured by reference is a nice way to make the live refresh possible without re-registering everyone, the upgrade path is safe (falls back to defaults when an older config lacks the key), and the broadcast → reload wiring hangs together. A few things worth tightening before it goes in.

**The watcher's idea of "excluded" doesn't quite match the explorer's.** `buildIgnoreList` ignores `**/<name>/**`, which only matches stuff *inside* a folder called `<name>`. But `readDir` and `searchFiles` exclude by exact-name set membership — any file or folder with that name, at any level, including the entry itself. So the explorer hides things the watcher still watches. It's the same glob that was there before, but since the PR is explicitly about unifying these three surfaces, this is the spot to make them actually agree — something like `[`**/${name}`, `**/${name}/**`]` would cover the entry and a same-named file too.

**The refresh swaps watchers without detaching the old one's listeners.** In `refreshWatcherIgnores` the new watcher is created and `old.close()` is called, but it's not awaited and the old listeners aren't removed first. During that close window both watchers are live and share the same subscribers map, so an event landing right then fires on both. IPC subscribers dedupe by path so they're fine, but in-process subscribers (the git monitor) have no coalescing and would fire twice. A quick `old.removeAllListeners()` before `close()` closes that gap.

**`SETTINGS_CHANGED` fires for every setting, not just exclusions.** Right now every `setSetting` broadcasts to all windows, and each FileExplorer runs its listener and string-compares the key. It's correct — nothing spurious reloads — but it means dragging e.g. the zoom-speed slider pings every window/explorer. Scoping the broadcast to the keys that actually have live listeners would avoid the churn.

**Smaller stuff:** the `onSettingsChanged` callback is typed `key: string` in the preload but `keyof AppSettings` in `electron-api.d.ts` — no runtime issue, just worth aligning. And the settings input rejects path separators but still accepts glob metacharacters like `*`/`?`; those are dead entries in `readDir` (literal match) but chokidar *does* interpret them in the ignore list, which is another small way the two surfaces diverge — might be worth rejecting them outright.

One logistics note: there's a `CHANGELOG.md` conflict that'll need resolving before this can merge.

None of this is blocking in principle — happy to see it land once the watcher/explorer semantics are reconciled (the first two points are the ones I'd really want addressed).

cbergoon:
@PaulHorn 

In regards to the difference in semantics on the expectations of the file patterns: 

Deliberate choice for now: one list for all three surfaces. The model matches the common case (you almost always want hide+ don't-search+don't-watch together), and after this PR the three surfaces are actually consistent. I'm aware of the node_modules "browse-but-don't-watch" gap. 

If it comes up I'd split into separate visibility/search/watch lists rather than overload this one, but I think its extra complexity for the user to understand. 

Single list is the intended near-term model, with a clear seam to split later if needed.


The changes on the last commit are summarized below: 

1. Watcher exclusions didn’t match the explorer

The explorer excludes files and folders by name (e.g. dist) regardless of where they appear. The watcher was only excluding the contents of excluded folders, not the folder itself or files with the same name. That meant the watcher could still see things the explorer was hiding. Updated the watcher exclusions so they match the explorer behavior exactly. If something is excluded, it’s excluded everywhere.

2. Watcher refresh could briefly double-fire events

When the exclusion list changed, we rebuilt the file watchers so the new rules took effect immediately. The problem was that the old watcher shuts down asynchronously, creating a short window where both the old and new watchers were active and could emit the same event twice. Fixed by detaching the old watcher before closing it and adding some error handling around watcher creation. This eliminates the duplicate-event window and prevents losing file events if a watcher fails to rebuild.

3. SETTINGS_CHANGED was firing too often

To support live exclusion updates, we broadcast a SETTINGS_CHANGED event so the explorer can refresh. The issue was that this event was firing for every setting change, even things like font size or zoom preferences. Added a whitelist of settings that actually require live updates. Right now that’s just fileExclusions, which cuts down on unnecessary refreshes and IPC traffic.

4. Type mismatch in the settings API

The preload implementation and the public API definition had drifted apart. One treated the settings key as a generic string while the other used keyof AppSettings. Aligned the types so both sides agree and TypeScript can catch issues correctly.

5. Exclusion names could be interpreted as globs

The explorer treats exclusion entries as literal names, but the watcher converts them into glob patterns. Special characters like *, ?, or [abc] could cause the watcher to exclude different files than the explorer. Added validation to reject glob metacharacters in exclusion names. Exclusions are now always treated as literal file or folder names across the entire app.

Cam

cbergoon:
Another thought on this, in the future, .gitignore could be factored in. This might make sense if it becomes problematic.

PaulHorn:
Thanks Cam — this all looks right to me, and I agree with the single-list call for now. The "browse-but-don't-watch" seam being explicit is exactly what I wanted documented, and the `.gitignore` idea is a sensible future direction if exclusions start to feel like duplicated bookkeeping.

I went through the latest commit point by point:

- **Watcher/explorer agreement** — `**/<name>` + `**/<name>/**` now matches `readDir`/`searchFiles` exact-name semantics (entry itself + same-named files + folder contents, at any depth). 👍
- **Refresh double-fire** — `removeAllListeners()` before the async `close()` closes the window where both watchers shared the subscribers map. Good catch on the in-process git monitor not coalescing.
- **Scoped broadcast** — `LIVE_REACTIVE_SETTINGS` whitelist means slider edits no longer ping every explorer. Clean.
- **Type alignment + glob rejection** — both done; literal-only names keep the three surfaces honest.

I pulled the branch and verified locally: full suite green (325 passing), main-process typecheck clean. I also added a small test (`fileExclusions.test.ts`) that pins the behavior we care about — `readDir`/`searchFiles` excluding by exact name at any depth (folder *and* same-named file), and the live per-call read so an edit takes effect with no relaunch. Pushed it to the branch.

One minor, non-blocking note for later: the glob-metachar rejection lives in the renderer input only — the main-side schema just validates `array`, so a hand-edited config or direct IPC could still slip a `*` into the watcher's ignore globs. Not worth gating this PR on; just flagging if we ever harden the settings boundary.

Merging once CI is green. Nice work.

PR #162, opened 2026-05-29
working draft / idea: Overview — global view of all workspaces

schroejahr2:
Bildschirmfoto 2026-05-29 um 09 55 41

Bildschirmfoto 2026-05-29 um 10 06 57

> **Draft / idea — work in progress.** Not ready for review yet.

Just to check acceptance / idea. 

Would be nice to have an overview. Screenshot attached. 

Working -> but just a shot, needs polishing, code hardening, styling, attention needed marker, rendering on zoom in, content preview.  etc.etc 

jumps to - and mounts terminal on click, overview is able to pan / zoom etc, sorted by workspace.

## What

A full-screen **Overview** overlay that shows every workspace at once: each workspace as a labeled, tinted box with all of its windows tiled at uniform size, in a zoom/pan-able space. Reachable from a new rail button below the workspace icon group.

Goal: a spatial bird's-eye over all workspaces + fast navigation (click a window → jump there).

## How it works (MVP)

- **Schematic cards**, no live panel mount: each window = title bar (icon + title), big centered logo, short snippet (editor→file, terminal→cwd, browser→url).
- **Agent logos**: terminal cards running a detected agent show the agent brand logo (Claude Code, Codex, …), from live agent status with a title fallback.
- **Two data sources** so inactive workspaces render too: visited workspaces from in-memory `canvasNodes`+`panels`; deferred (never-opened) ones from the lazy session snapshot.
- **Navigation**: click a card → switch workspace + focus/center the node (visited only — ids regenerate on deferred restore); click the box background → switch workspace only.
- **Zoom + pan**: Cmd+scroll = zoom, two-finger = pan, Esc closes. Initial fit-to-screen.

## Out of scope (phase 2)

Live panel content on zoom-in (real terminal/editor) — needs a `renderPanelContent` refactor (currently bound to the active workspace) + WebGL context limits.

## Files

- New: `src/renderer/overview/{collectOverview,navigate}.ts`, `{OverviewMode,WorkspaceBox,OverviewCard}.tsx`
- Wiring: `uiStore.showOverview`, `App.tsx`, rail button in `Sidebar.tsx`, `selectAgentInfoByPanel` exported from `useAgentPanelInfo`

## Status

Typecheck + build clean; existing test suite green (minus the documented env-only git tests). Manually clicked through locally. No tests for the new feature yet.

###  Feedback

what do you think about this ?

### DIFF
--- a/src/renderer/App.tsx
+++ b/src/renderer/App.tsx
@@ -26,6 +26,7 @@ import { renderPanelComponent, PANEL_REGISTRY } from './panels/registry'
 const CanvasPanel = PANEL_REGISTRY.canvas.Component
 import { NodeSwitcher } from './ui/NodeSwitcher'
 import { CommandPalette } from './ui/CommandPalette'
+import OverviewMode from './overview/OverviewMode'
 import { SettingsWindow } from './settings/SettingsWindow'
 import { SavedLayoutsDialog } from './dialogs/SavedLayoutsDialog'
 import { PostUpdateFeedbackDialog } from './dialogs/PostUpdateFeedbackDialog'
@@ -115,6 +116,7 @@ function MainApp() {
   const selectedWorkspaceId = useAppStore((s) => s.selectedWorkspaceId)
   const showNodeSwitcher = useUIStore((s) => s.showNodeSwitcher)
   const showCommandPalette = useUIStore((s) => s.showCommandPalette)
+  const showOverview = useUIStore((s) => s.showOverview)
 
   // Theme — apply on mount and re-apply whenever appearanceMode changes
   const appearanceMode = useSettingsStore((s) => s.appearanceMode)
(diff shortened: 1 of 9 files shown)


PaulHorn:
Interesting idea, and I like the direction. We actually had something a bit like this in Cate before, but it was exposed as a shortcut rather than its own mode, and I think that framing fits better here too. A few things I'd want to see before it moves out of draft:

1. Make it shortcut driven instead of a sidebar button. Rather than a permanent activity bar entry, I'd open the overview with a keyboard shortcut and let you tab through the workspaces from there. That matches how it was used before and keeps it light.

2. Drop the German UI strings. The labels and tooltips ("Übersicht", "Schließen (Esc)", and so on) should be English so it matches the rest of the app.

3. Rethink the layout. Stacking the workspaces straight below each other doesn't really scale. I'd love to see a smart grid that always fills the viewport, arranging the boxes to use the space they have instead of one long column.

Good input overall, keen to see where this goes once it's reworked a bit.

PR #175, opened 2026-05-29
fix(terminal): send macOS line-editing key sequences to the shell

### LINKED ISSUE #172: Terminal: Cmd+Delete (and macOS line-editing keys) don't reach the shell
## Bug

In a Cate terminal, **`Cmd+Delete` (`Cmd+Backspace`) does not delete to the start of the line** the way it does in the VS Code / Cursor integrated terminal. Standard macOS line-editing chords are missing or wrong (e.g. `Option+Backspace` does not delete a word, `Cmd+←/→` don't jump to line start/end).

## Root cause

Two layers both fail to deliver the chord to the shell:

1. **App layer eats it.** `deleteNode` is bound to `Cmd+Backspace` (`src/shared/types.ts:506`). The capture-phase keydown handler in `src/renderer/hooks/useShortcuts.ts:275-277` deliberately runs `deleteNode` even when a terminal is focused, calling `preventDefault()`/`stopPropagation()` — so the keystroke never reaches xterm. (It tries to close the focused panel instead.)
2. **Terminal layer never translates it.** `src/renderer/lib/terminalRegistry.ts:606` returns `true` for any `metaKey` combo, so xterm's default fires — a plain `\x7f` (delete one char) for `Cmd+Backspace`. `Option+Backspace` hits the CSI-u branch and sends `\x1b[127;3u`, which a plain shell (readline/zsh) doesn't interpret as delete-word.

## How VS Code does it (reference)

VS Code registers keybindings gated by `when: terminalFocus` -> `workbench.action.terminal.sendSequence`, writing literal bytes to the PTY. Verified macOS defaults (`terminalContrib/sendSequence/.../terminal.sendSequence.contribution.ts`):

| Chord (mac) | Bytes | Effect |
|---|---|---|
| `Cmd+Backspace` | `\x15` (Ctrl+U) | delete to line start |
| `Option+Backspace` | `\x17` (Ctrl+W) | delete word left |
| `Option+Delete` (forward) | `\x1bd` (ESC d) | delete word right |
| `Cmd+Left` | `\x01` (Ctrl+A) | line start |
| `Cmd+Right` | `\x05` (Ctrl+E) | line end |
| `Option+Left` | `\x1bb` (ESC b) | word left |
| `Option+Right` | `\x1bf` (ESC f) | word right |

## Proposed fix

- New pure module `terminalKeymap.ts` with a data-driven table + `resolveTerminalKeySequence(event, isMac)` (macOS-only for now; extensible to Win/Linux).
- Translate in the xterm `customKeyEventHandler` (works for canvas, dock, and detached-window terminals — unlike an app-layer `terminalFocus` check which is canvas-centric). Extract the duplicated handler in `getOrCreate`/`reconnectTerminal` into one factory.
- `useShortcuts`: stop consuming `Cmd+Backspace` when a text surface (incl. terminal) is focused. Closing a focused terminal panel stays on `Cmd+W`.

## Expected behavior

The 7 chords above behave exactly as in VS Code/Cursor when a Cate terminal is focused. CSI-u disambiguation for `Enter`/`Tab`/`Escape`/`Space` is unchanged.

Related: overlaps the keyboard-rework scope of #167. Design doc: `docs/superpowers/specs/2026-05-30-terminal-macos-line-editing-keys-design.md`.

architawr:
## What & why

In Cate terminals, **`Cmd+Delete` (`Cmd+Backspace`) didn't delete to the start of the line**, and other macOS line-editing chords were missing/wrong — unlike the VS Code / Cursor integrated terminal. Two layers were at fault:

1. The app shortcut handler ran `deleteNode` on `Cmd+Backspace` even when a terminal was focused, calling `preventDefault()` before xterm saw the key.
2. The terminal's `customKeyEventHandler` returned `true` for any `Cmd` combo and never translated it to a readline control sequence.

## Changes

- **New `src/renderer/lib/terminalKeymap.ts`** — a pure, data-driven table + `resolveTerminalKeySequence(event, isMac)` mapping the macOS chords to the exact bytes VS Code's integrated terminal sends. macOS-only for now; extensible to Win/Linux by adding rows.
- **`terminalRegistry.ts`** — translate the chords in the xterm key handler (works for canvas, dock, and detached-window terminals). Extracted the duplicated handler from `getOrCreate`/`reconnectTerminal` into one shared `createTerminalKeyEventHandler` factory so they can't drift apart. The CSI-u path for `Enter`/`Tab`/`Escape`/`Space` is unchanged.
- **`useShortcuts.ts`** — `deleteNode` no longer fires when a text surface (incl. a focused terminal — xterm focuses its `.xterm-helper-textarea`) has focus, so `Cmd+Backspace` reaches the shell. A focused terminal panel is still closed with **`Cmd+W`**. `undo`/`redo` behaviour is unchanged.

### Chords (mac, VS Code parity)

| Chord | Bytes | Effect |
|---|---|---|
| `Cmd+Backspace` | `\x15` (Ctrl+U) | delete to line start |
| `Option+Backspace` | `\x17` (Ctrl+W) | delete word left |
| `Option+Delete` | `\x1bd` (ESC d) | delete word right |
| `Cmd+Left` / `Cmd+Right` | `\x01` / `\x05` | line start / end |
| `Option+Left` / `Option+Right` | `\x1bb` / `\x1bf` | word left / right |

## Testing

- New `terminalKeymap.test.ts` (13 cases): every chord → expected bytes, non-mac → null, plain/near-miss combos → null.
- Full suite: **333 passed, 3 skipped, 0 failed**. `npm run build` ✓. `tsc --noEmit` ✓.
- `useShortcuts` is a store-coupled hook with no existing unit tests; its change is a single guard branch, verified via build/typecheck.

Closes #172

### DIFF
--- a/src/renderer/hooks/useShortcuts.ts
+++ b/src/renderer/hooks/useShortcuts.ts
@@ -269,11 +269,19 @@ export function useShortcuts(): void {
       // Context-aware guard: when a text surface (input, textarea, Monaco,
       // xterm helper textarea, contenteditable) has focus, let clipboard and
       // undo/redo fall through to it natively instead of the canvas.
-      // Terminals don't consume Cmd+Z/Y/Backspace, so let canvas handle them
-      // even when a terminal panel is focused. Only real text editors
-      // (Monaco, inputs, contenteditables) should swallow them.
-      if (action === 'undo' || action === 'redo' || action === 'deleteNode') {
+      // Terminals don't consume Cmd+Z/Y, so the canvas handles undo/redo even
+      // when a terminal is focused — only real editors/inputs swallow them.
+      if (action === 'undo' || action === 'redo') {
         if (!terminalHasFocus && isTextSurfaceFocused()) return
+      }
+      // Cmd+Backspace (deleteNode): a focused terminal now owns it for line
+      // editing (kill-to-line-start, sent to the shell by the terminal's key
+      // handler), so never delete a focused terminal/editor/input panel with
+      // it. The xterm helper textarea counts as a text surface, so this covers
+      // canvas, docked, and detached-window terminals alike. Close a focused
+      // terminal panel with Cmd+W instead.
+      if (action === 'deleteNode') {
+        if (isTextSurfaceFocused()) return
       }
 
       // Keyboard-only passthrough: when a browser panel is focused, let
--- a/src/renderer/lib/terminalKeymap.test.ts
+++ b/src/renderer/lib/terminalKeymap.test.ts
@@ -1,0 +1,92 @@
+import { describe, it, expect } from 'vitest'
+
+import { resolveTerminalKeySequence, MAC_TERMINAL_KEYMAP } from './terminalKeymap'
+
+/** Build a minimal KeyboardEvent-shaped object. The resolver only reads
+ *  key + the four modifier flags, so a plain literal is enough (no jsdom). */
+function ev(
+  key: string,
+  mods: Partial<Record<'metaKey' | 'altKey' | 'ctrlKey' | 'shiftKey', boolean>> = {},
+): KeyboardEvent {
+  return {
+    key,
+    metaKey: false,
+    altKey: false,
+    ctrlKey: false,
+    shiftKey: false,
+    ...mods,
+  } as unknown as KeyboardEvent
+}
+
+describe('resolveTerminalKeySequence — macOS line-editing chords (VS Code parity)', () => {
+  it('Cmd+Backspace → kill to line start (Ctrl+U)', () => {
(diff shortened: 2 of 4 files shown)


PaulHorn:
Thanks for this @architawr — solid fix, and the writeup in #172 made it easy to verify. We landed the same change in #177, which targets current `main`. This branch was cut before #174 merged, so it still imports the old `getResolvedTheme`/`ResolvedTheme` theme API that #174 removed (hence the conflicts) — merging as-is risked reverting that theming refactor.

#177 keeps your design intact: it adopts your data-driven `MAC_TERMINAL_KEYMAP` table (key/meta/alt/send/label) and the table test, rebased onto the post-#174 terminalRegistry. Closing this in favour of #177 — much appreciated. 🙏

PR #176, opened 2026-05-29
feat(tabs): close tab on middle-click

architawr:
## What & why

Middle-clicking a tab did nothing. This adds **middle-click-to-close**, matching browsers and VS Code — for both the **top dock tab bar** and **canvas-node windows** (terminal etc.). Both render the same `TabPill`, so a single-tab window on the canvas closes with one middle-click too.

Per discussion, scope is **tabs only** — there is no "middle-click the header frame to nuke a whole multi-tab window" gesture (predictable, nothing closes by accident).

## Changes

- New `src/renderer/lib/mouse.ts` — `isMiddleClick(e)` predicate. `auxclick` fires for **both** the middle (1) and right (2) buttons, so this guards close-on-middle so **right-click still opens the context menu**.
- `DockTabBar.tsx` — `onAuxClick` on the tab closes it via the existing `onClosePanel`; middle-button `onMouseDown` is `preventDefault`-ed to suppress the native autoscroll affordance and avoid starting a tab drag.

## Testing

- New `mouse.test.ts`: button 1 → true, buttons 0/2 → false (locks the right-click guard).
- Full suite: **323 passed, 3 skipped, 0 failed**. `npm run build` ✓. `tsc --noEmit` ✓.
- The `onAuxClick` wiring sits on the heavily store/registry-coupled `DockTabBar`; the testable decision (`isMiddleClick`) is unit-tested, the wiring verified by build + typecheck.


Anton-Horn:
Good change, well scoped. Will merge once CI ran through. Thanks!

PR #178, opened 2026-05-29
feat(terminal): clickable links (Cmd/Ctrl+Click) like VS Code

architawr:
## What & why

Terminal URLs weren't highlighted or clickable. This adds **clickable links** matching VS Code / Cursor: hover underlines a URL, and:

- **Cmd+Click** (mac) / **Ctrl+Click** (win/linux) → opens in an **in-app BrowserPanel**, reusing the existing auto-open routing + dedupe (`openTerminalUrl`) — so it reuses/creates a browser panel and won't double-open a URL already surfaced by auto-open.
- **Cmd+Shift+Click** / **Ctrl+Shift+Click** → opens in the **external system browser** (`electronAPI.openExternalUrl`).
- plain click → ignored (same as VS Code).

## Changes

- Dependency: `@xterm/addon-web-links@^0.11.0` (official xterm addon; hover-underline + http/https detection).
- New `src/renderer/lib/terminalLinks.ts` — pure `resolveTerminalLinkTarget(e, isMac)` returning `'panel' | 'external' | 'ignore'` (primary modifier = Cmd on mac / Ctrl elsewhere; +Shift → external).
- `terminalRegistry.ts` — loads `WebLinksAddon` in both the fresh-spawn and reconnect paths via one shared `createTerminalLinkHandler(workspaceId)` factory.

## Scope

http/https only. Bare `localhost:3000` (no scheme) stays with the existing auto-open; **file-path → editor** links are a separate follow-up PR.

## Testing

- New `terminalLinks.test.ts`: the full modifier × platform matrix (8 cases) — incl. Ctrl-on-mac and Cmd-on-non-mac → `ignore`.
- Full suite: **328 passed, 3 skipped, 0 failed**. `npm run build` ✓. `tsc --noEmit` ✓.
- The `WebLinksAddon` wiring is glue, verified by build + typecheck.

PR #180, opened 2026-05-29
feat(terminal): open file paths in the editor on Cmd+Click

architawr:
## What & why

File references in terminal output (tsc/eslint/test lines like `src/foo.ts:12:5`) are now **underlined** when they point at a real file and, on **Cmd+Click** (mac) / **Ctrl+Click** (win/linux), **open in an editor jumped to the line/column** — like VS Code. Companion to the URL clickable-links PR (#178); this one handles **files**.

## Changes

- New `terminalFileLinks.ts` (pure): `parseTerminalFileMatches` (path + optional `:line:col`, slash-or-absolute required), `resolveCandidatePath` (relative → workspace/cwd root), `isTerminalLinkModifier`.
- New `editorReveal.ts` (pure): one-shot `setPendingReveal`/`takePendingReveal` — set right after `createEditor`, consumed once by `EditorPanel` after its Monaco model loads. No `PanelState`/persisted-state changes, no re-jump on remount.
- New `terminalFileLinkProvider.ts` (glue): xterm `ILinkProvider` that reads the buffer line, validates candidates via `fsStat` (cached), and opens existing files at the line on the modifier click. URLs handled separately by WebLinksAddon.
- `terminalRegistry.ts`: `registerLinkProvider(...)` in both the fresh-spawn and reconnect paths; disposable tracked in `cleanupListeners`.
- `EditorPanel.tsx`: applies the pending reveal (`revealLineInCenter` + `setPosition` + focus) once, after the model is set.

## Scope

- Opens a **new** editor (matches the app's existing open-file behaviour — no dedupe).
- **Only existing files** are linkified (`fsStat`), so prose and dead paths aren't highlighted.
- No bare-filename (no-slash) matching; relative base is the terminal **cwd / workspace root** (not the live PTY cwd).

## Testing

- New tests: `terminalFileLinks.test.ts` (15 — parsing matrix, resolve, modifier), `editorReveal.test.ts` (4 — one-shot semantics).
- Updated `terminalRegistry.test.ts` mock to mirror the new `registerLinkProvider` + `useAppStore` surface the code uses.
- Full suite: **339 passed, 3 skipped, 0 failed**. `npm run build` ✓. `tsc --noEmit` ✓.


Anton-Horn:
Good follow up on #178; Will merge once CI ran through. Thanks!

PR #185, opened 2026-05-29
feat(terminal): add a scroll-speed setting (xterm scrollSensitivity)

architawr:
## What & why

Terminal vertical wheel-scroll felt **much faster than canvas scroll**, and there was a canvas **zoom-speed** setting but **no terminal scroll-speed** one.

Root cause: xterm was constructed with only `scrollback` set — `scrollSensitivity` stayed at its default `1.0` and scrolled the viewport natively/un-throttled, while canvas pan is rAF-batched (and zoom is lerped + multiplied by `zoomSpeed`). The terminal had no speed knob at all.

## Change

Add a **`terminalScrollSpeed`** setting mapped to xterm's purpose-built `scrollSensitivity`, mirroring the existing `zoomSpeed` wiring end to end:

- `src/shared/types.ts` — `terminalScrollSpeed: number` in `AppSettings` + `1.0` in `DEFAULT_SETTINGS`.
- `src/main/store.ts` — `terminalScrollSpeed: 'number'` in `SETTINGS_SCHEMA`.
- `src/renderer/lib/terminalRegistry.ts` — pure `clampScrollSensitivity()` (0.25–3.0; invalid/≤0/NaN → 1.0) + `getScrollSensitivity()`; set in **both** Terminal constructors (getOrCreate + reconnectTerminal); `applyScrollSensitivityToAll()` wired into the existing settings subscriber so open terminals update **without reload** (same pattern as cursor-blink).
- `src/renderer/settings/TerminalSettings.tsx` — a "Scroll speed" `Slider` (0.25–3.0, step 0.25), like the canvas zoom-speed slider.

## Behavior

- **Default 1.0 = no change** for existing users. Set 0.25–0.5 to slow the terminal toward canvas feel.
- Applies to all terminals (canvas/agent/docked/detached — shared registry).
- Canvas pan left untouched (intentionally 1:1 + inertia; px vs rows are different units, so this makes terminal scroll *adjustable*, not byte-identical).

## Testing

- New unit tests for `clampScrollSensitivity` (bounds, clamp up/down, invalid → 1.0).
- Full suite **369 passed / 3 skipped / 0 failed**. `tsc --noEmit` ✓. `npm run build` ✓.
- Adversarially reviewed (multi-agent): 24 findings raised, 23 refuted on verification, 0 real defects.


Anton-Horn:
Good PR. Will merge after CI passed. Thanks!

PR #193, opened 2026-05-30
feat(sidebar): per-tab rename + context menu (Rename / Close) from the sidebar

architawr:
## Why

In the left sidebar, each workspace expands to list its individual tabs/panels as rows. But the only rename/context-menu plumbing lived on the `WorkspaceTab` root `<div>`, wrapping both the workspace header and every panel row. Right-clicking (or double-clicking) a specific panel row bubbled up to that single workspace-level handler, which always acted on `workspace.id` — so renaming from a panel row renamed the WHOLE workspace, never the tab the user clicked.

## What

Each panel row now gets its own rename trigger that targets just that tab:
- **Right-click on a panel row** opens a dedicated "Rename" context menu. The handler calls `e.preventDefault()` + `e.stopPropagation()` so it does not fall through to the workspace context menu.
- **Double-click on the row label** starts an inline rename, mirroring the workspace-header UX for consistency.
- The inline `<input>` reuses the workspace rename pattern: Enter / blur commits, Escape cancels, click is swallowed so it doesn't trigger the row's focus-panel handler.
- Rename state is tracked per-panel (`renamingPanelId` / `panelRenameValue`), distinct from the workspace `isRenaming` state, and rendered in place of the active row's label.
- Commits via the existing `renamePanelByUser(workspace.id, panelId, trimmedTitle)` action (sets `panel.title` + `titleUserOverridden`). Empty/whitespace-only input is ignored, matching the workspace rename. The workspace name is left untouched.

This works for both row variants: the `TerminalPanelRow` component (terminal/agent panels) and the generic panel `<button>` (editor/browser/git/etc.). No store changes.

## Files

- `src/renderer/sidebar/WorkspaceTab.tsx` — add `PanelRenameProps` + `PanelRenameInput`; thread per-panel rename state, `beginPanelRename` / `handlePanelRenameSubmit` / `handlePanelContextMenu` callbacks; wire double-click + context-menu + inline input into both row types.
- `src/renderer/sidebar/WorkspaceTab.test.tsx` — focused `TerminalPanelRow` rename-wiring tests: double-click begins rename, right-click routes to the panel context-menu handler (not the workspace), inline input commits on Enter, Escape cancels.

## Testing

- `npx tsc --noEmit` — clean.
- `npm test` — 42 files, 415 passed / 3 skipped (skips pre-existing).

---

### Update — panel-row context menu

The per-panel right-click menu now mirrors the dock tab menu with the actions that apply to a flat sidebar list:

- **Rename** — inline-rename this panel (as before).
- **Close** — closes the specific panel via `closePanel(workspaceId, panelId)` (disposes terminal / releases canvas store / removes from dock or canvas).

Dock-stack-relative items (Split Right, Close Others, Close to the Right, Move into New Window) are intentionally omitted — they have no meaning for the sidebar's flat list.

Verified: `tsc --noEmit` clean; `npm test` 415 passed / 3 skipped.

### Update — middle-click to close

Middle-click (auxclick button 1) on a sidebar panel row now closes that panel, mirroring the dock tabs. `onMouseDown` suppresses the native middle-click autoscroll, and the middle-button guard keeps right-click opening the context menu. Reuses the shared `isMiddleClick` helper; covers both row variants. +2 row-level tests (middle-click closes; right-click does not). Suite: 417 passed / 3 skipped.


architawr:
@PaulHorn — ready for review when you get a chance 🙏

PR #211, opened 2026-05-30
feat(agent): let the Cate agent control panels (cate-control)

architawr:
## Summary

Adds **cate-control**: the in-app Cate agent (pi) can now drive the workspace itself — open, arrange, focus/move/resize/close panels and interact with their contents (run a terminal command, navigate a browser, reveal a file at a line, toggle markdown preview) on the canvas it lives in. The agent operates the *workspace* (what's open, where it sits, what's running, where the camera looks); file edits stay on its normal `Edit` tool.

12 new `cate_*` tools, a per-chat Guarded/Auto toggle, and a global on/off setting.

## How it works

**Transport.** pi exposes no generic host RPC — only `ctx.ui.select/confirm/input`. So a bundled `cate-control` pi extension piggybacks a structured request on `ctx.ui.input()` with a sentinel-prefixed JSON payload (`@@cate-control@@{...}`). The renderer intercepts the sentinel in `agentStore.handleEvent` **before** the dialog queue, dispatches it, and replies over the existing `agent:uiResponse` channel. Pure-visual actions (pan/zoom) are fire-and-forget.

**Dispatcher.** `src/agent/renderer/cateControl.ts` resolves the calling chat's workspace/canvas via a context registry that `AgentPanel` populates (`useCanvasStoreApi()`), classifies the action (safe vs side-effect), applies the chat's mode policy, runs an executor (`cateExecutors.ts`), and returns a structured response.

**Modes (guarded / auto).** A per-chat toggle in the agent chat footer (next to plan-mode), mirroring plan/auto in Claude Code:
- *Guarded* (default) — reads + safe panel ops run immediately; side-effects (close, run command, open URL) prompt an inline Allow/Deny card (reuses the existing approval UI).
- *Auto* — everything runs without prompting.

A global `cateControlEnabled` setting gates the whole feature.

**Semantic placement.** The agent never sends pixels — it states intent (`right of X`, `tile`, `relativeTo: 'self'`); `cateControlLayout.ts` (pure geometry, unit-tested) computes canvas-space rects.

**Self-protection.** `cate_get_layout` marks the agent's own host panel `isSelf: true`; executors refuse to close/move it and exclude it from `arrange` unless targeted explicitly by id. The feature is placement-agnostic so a future "global/operator agent" can land without rework.

## Tool surface (12 `cate_*` tools)

- **Query:** `cate_get_layout`
- **Lifecycle:** `cate_open_panel` (editor/terminal/browser/git/fileExplorer/document), `cate_close_panel`
- **Management:** `cate_focus_panel`, `cate_move_panel`, `cate_resize_panel`, `cate_arrange` (tile/grid/cascade/focus-one)
- **Content:** `cate_run_in_terminal`, `cate_open_url`, `cate_reveal_in_editor`, `cate_set_markdown_preview`
- **Viewport:** `cate_pan_to`, `cate_zoom`

## Two fixes from live testing

- **Terminal commands now actually run.** `createTerminal`'s `initialInput` is intentionally not persisted (it would re-run on session restore), so `open_panel(terminal, command)` silently dropped the command; `run_in_terminal`'s single 250 ms retry was too short for a fresh `node-pty` to register. Both now route through `writeToTerminalWhenReady`, which polls the terminal registry until the PTY is live, then writes the command.
- **Markdown preview is reachable.** The app already supported preview (`EditorPanel` → `setPanelMarkdownPreview`) but the agent had no tool. Added `cate_set_markdown_preview` + a `preview` option on `cate_reveal_in_editor`.

## Implementation notes

- Bundled extension installs per-workspace on agent-session start (mirrors `cate-plan-mode`); shipped in prod via the existing `extraResources` glob.
- UI change: one small Guarded/Auto toggle button in the agent chat footer (both sidebar + dock render sites).
- Registered `cateControlEnabled` in `SETTINGS_SCHEMA` (`src/main/store.ts`) so `tsc` stays clean.

## Testing

- **Unit (Vitest):** classifier, pure placement geometry, dispatcher gating, per-chat mode state, and all executors (incl. condition-based terminal send + preview). Full suite **453 passing**; `tsc --noEmit` clean; `npm run build` green.
- **E2E (Playwright — `e2e/cate-control.spec.ts`):** drives the real renderer dispatcher in a live Electron window — `run_in_terminal` and `open_panel(terminal, command)` actually execute in a spawned PTY (asserted via command output in the xterm buffer), and `set_markdown_preview` flips the editor into preview.
- **Manual:** verified live (`npm run dev`) — toggle renders + flips, panels open/arrange, terminal commands run, preview toggles.

## Future ideas (brief)

- **Drag a panel into the chat** to add it to context (file / terminal output / browser) — unique to the spatial model.
- **Visual session tree** — render pi's branch tree as canvas nodes.
- **Render `ctx.ui.custom` extensions** as native panels — unblocks much of the pi package catalog (currently flagged "requires terminal").
- **Global / operator agent** — lift the agent out of the canvas as an app-level driver (this work is deliberately placement-agnostic to allow it).
- **Per-task model routing**, **multi-workspace control** (v1 is active-workspace + main-window only), and **richer permission rules** for unattended Auto mode.

## Test plan

- [x] `npm run build` passes
- [x] `npm run test` passes (453)
- [x] In-app: toggle Guarded/Auto; ask the agent to open a file, run a terminal command, tile panels, toggle markdown preview
- [x] Guarded mode prompts Allow/Deny for side-effects; Auto runs without prompts


Anton-Horn:
We have similar behaviour implicit by letting any agent write to the workspace.json file. Not quite the same but it fits in more into the overall application instead of placing the cateControl system on top. I do see the advantages and I like it being a pi extension. What I don't like so much about it: I don't think the agent should control zoom/camera position. Additionally I noticed when creating a new panel instead of focusing that one it just pan's to a random location. 

two things I would like to see here: 
1. make it a bit more basic: remove use actions like panning and zooming. Add in a terminal read (so this is not just a canvas edit feature but also supports agent orchestration). Make sure the tools are all optimised for agents (keep them lean and focused, maybe reduce count, 12 feels very heavy)

2. include custom tool renderings. This is quite a "flashy" feature so we should display it as it. (custom tool rendering's, not just displaying raw json, approval workflows ). Make sure it fit's to the general UI of the agent panel.

Anton-Horn:
Removed the skill.md for workspaces in #214  so this one is supposed to be the new way how agents can control the workspace/do orchestration. Will be merged once the changes are in and the feature is polished. Thanks!

architawr:
Addressed both points:

**1. Leaner, more agent-focused toolset (13 → 11).**
- Removed `pan_to` (it was literally identical to `focus_panel` — both just `focusAndCenter`) and `zoom` (the agent shouldn't drive zoom/camera).
- Folded `reveal_in_editor` into `open_panel`: `open_panel` now focuses + centers what it opens and accepts `target.preview` for markdown, so the dedicated reveal tool was redundant.
- Added `read_terminal` — reads a terminal panel's recent buffer (screen + scrollback) as text, so the agent can inspect output it ran via `run_in_terminal` (the other half of orchestration).
- Also fixed the "new panel pans to a random location" bug: `open_panel` no longer estimates the viewport as the centroid of all nodes — it centers on the real viewport (`containerSize` + `viewToCanvas`) and focuses the panel it opened.

**2. Custom tool renderings + approval workflow.**
- cate-control calls now render in the thread as compact, accent-tinted cards (icon + verb + summary, expandable to params/result) with running → success/denied/error status — instead of being silent round-trips.
- The guarded-mode approval card renders the action with an icon + human-readable request ("Let Cate run `npm test`?") rather than `cate:<action>` + a JSON dump.
- A shared `cateToolDisplay` maps `(action, params) → { icon, verb, summary }` so the thread card and approval card stay consistent.

Merged latest `main`; `typecheck` / `test` / `build` all green.

architawr:
@PaulHorn — would appreciate your eyes on this one too when you have a moment 🙏

Anton-Horn:
I'll check it later. 11 tools still seem like a lot of complexity / token usage. Will give you more detailed feedback then. Thanks for the update.

architawr:
@Anton-Horn — update on the complexity / token-usage point.

**Consolidated 11 → 4 tools.** The agent now sees only:
- `cate_layout {op: get|arrange}` — read the canvas / rearrange panels
- `cate_panel {op: open|focus|move|resize|close|preview}` — single-panel lifecycle
- `cate_browser {panelId?, url}` — navigate a browser panel (room to grow: reload/back/JS)
- `cate_terminal {op: run|read}` — run a command / read its output

Thin op-routers delegate to the same focused executors, so per-op behavior and the self-protection guards (won't close/move the agent's own host panel) are unchanged. `classifyCateAction` still escalates only destructive (`close`) and outbound (run a command, navigate to a remote url) ops to guarded-mode approval; reads/focus/layout stay safe.

Custom rendering carries over from the earlier round: cate actions render as compact accent cards in the thread (icon + verb + summary, expandable to params/result), and guarded side-effects show a human-readable “Let Cate run `…`?” prompt instead of raw JSON.

**Also fixed a bug found while testing this:** the extension is copied into each workspace's pi-agent dir, and the installer used skip-if-exists — so after the toolset changed, the agent kept loading the stale copy and emitted action names the renderer no longer handled (“Unknown or unimplemented action”). Switched to refresh-on-change (`copyIfChanged`); this also fixes the latent case where a shipped extension update would never reach users who already had a copy.

`typecheck` / `test` (541) / `build` all green. Out of draft now — ready for your detailed look.

Anton-Horn:
I'm on it now. Would like to push some code and steer this pr myself. If that's fine with you.

architawr:
Ok, let's go

Anton-Horn:
updated the tool rendering and changed the tools schemas. What do you think?

architawr:
@Anton-Horn looks great overall — thanks for taking it forward. A few things I really like, and a few I want to confirm before we call it done.

**Like:**
- **The active-tools toggle (`/cate-on` / `/cate-off`).** Registering the tools but flipping them in/out of the active set is a much cleaner answer to the token-cost concern than a hard install/uninstall — the agent pays nothing for the definitions when it's off. 👍
- **Stable short-UUID panel ids.** Deriving the handle from the panel's own UUID (and surviving restore) instead of the parallel p1/p2 counter is the right call — and the exact-id → 6-char-prefix → title resolution with an ambiguous-prefix error is a nice touch.
- **Browser tool expansion.** `info`/`read`/`eval`/`screenshot` turn this into real orchestration rather than just "point at a URL."

**Want to confirm — intentional, or trimmed for the checkpoint?**
1. **Approval workflow removed.** The guarded/auto classifier + inline Allow/Deny card are gone, so everything now runs immediately under a plain on/off. I'm fine treating the in-canvas agent as trusted (same posture as its native tools), but it's worth being deliberate: the surface actually got *more* powerful here — `terminal run`, `browser eval` (arbitrary JS in the page), and `close` all execute with no gate. Do you want to keep it fully ungated, or re-add a light prompt for just the destructive/outbound ops (`close` / `terminal run` / `browser eval`)? Either way is fine by me — I just want it to be a decision, not a side effect of the refactor.
2. **`arrange` (tile/grid/cascade) dropped from `cate_layout`.** That was one of the headline "tile the panels for me" capabilities. Cut for scope, or coming back? The agent can still `move` panels one at a time, but there's no auto-layout now.
3. **`focus` / `resize` / preview-toggle dropped from `cate_panel`.** `open` focuses + centers what it opens, but there's no way to re-focus, resize, or toggle preview on an *already-open* panel. Intentional?

Tiny nits: the `cateToolDisplay.ts` header comment still says the agent addresses panels "by title" (now it's by id), and the `request` / "approval prompt" field is a leftover from the gated flow — worth cleaning up if approval stays gone.

Happy to pick any of these up if you'd rather I take them.

Anton-Horn:
This state feels solid but i don't think it's quite useful yet. Easier to just open a few terminals by myself then tell the agent too. Maybe we could move this into a orchestration direction but then i'm not sure if this is any better over subagents. Any idea how we could actually make this useful instead of just a gimmick?

architawr:
Totally agree — as a manual panel remote it's a gimmick. "Open three terminals" is faster to do myself, and pure orchestration is just subagents with extra steps. The thing it has that subagents *don't* is the canvas itself, so the useful version leans into that rather than into "agent clicks buttons." Three angles I think are actually worth it:

**1. Observable parallel work (this is the strongest one, IMO).** The value isn't the agent opening a terminal — it's the agent *laying work out in space* so you can watch and intervene. Kick off N parallel tasks → each in its own terminal, auto-tiled, with the relevant file/diff opened next to it. A subagent is a black box you read as a text log after the fact; here it's a live workspace you can glance at and grab any panel mid-run. This is exactly what `arrange` + a terminal-per-task pattern buys us — which is partly why dropping `arrange` made it feel less useful. So less "agent that uses Cate," more "Cate visualizes what the agent is doing."

I do have one reservation here: at some point this risks reimplementing what Claude Code / Codex already do (orchestration, parallel tasks), and we don't want to chase that. But the difference is that Cate *already ships its own agent and a harness built around it* — so rather than duplicating an external tool, we'd be adding capabilities to the agent we already own, on the one surface (the canvas) those external tools don't have. That's where it stops being a copy and starts being the thing only Cate can do.

**2. Drag-to-context (canvas → agent, the reverse direction).** Drag a terminal / browser / file panel into the chat to add its live contents as context. This is the bit that's genuinely impossible without the spatial model, and it makes the feature bidirectional instead of one-way "agent drives." Cheap to prototype and immediately useful day-to-day.

**3. Browser-as-a-real-tool.** `navigate`/`read`/`eval`/`screenshot` is basically built-in browser-use on the canvas: "open the staging dashboard, check the deploy landed, screenshot it." Subagents can't do that out of the box, and the result is something you can actually see on the canvas.

My honest read: #1 is what turns it from gimmick into "oh, I'd leave this on" — it reframes the whole feature as *observability for agent work* rather than a panel API. #2 is the highest value-per-effort. Want me to prototype the observable-orchestration flow (a task → tiled terminal + diff layout) on top of your current schemas so we can feel whether it lands? Happy to also just sit on it if you'd rather drive the direction.

Anton-Horn:
WIll close this for now. Don't see the value yet.

PR #228, opened 2026-05-31
fix(e2e): hide windows and disable animations under Playwright

architawr:
## What

Stop the e2e suite from opening windows on the user's active screen (clickable, focus-stealing) on every spec. Under `CATE_E2E` the windows are now never mapped to a display, and the renderer is made deterministic so the timing-sensitive drag specs stay reliable without a visible window.

## Why

Each Playwright spec launches the app, whose windows `show()` on macOS — which puts them on screen and **activates** the app. Because `CATE_E2E` runs from the same `node_modules/electron` (`Electron.app`) bundle as a running `npm run dev`, activating it also pulls the **dev instance's windows** forward. The windows appear in front of you, take focus, and register clicks.

Empirically there is **no window-positioning trick on macOS that is both invisible and preserves animation timing**:

| approach | drag/perf fails (baseline 4) | visible? |
|---|---|---|
| `show()` (baseline) | 4 | yes, steals focus |
| `showInactive()` | 3 | **yes** |
| off-screen `setPosition(-12000)` | 9 | **yes — macOS clamps it back to [0,34]** |
| hidden (never show) | 6 | no |
| **hidden + animations disabled** | **4** | **no** |

A shown window is kept on a display; a hidden window throttles rAF, so the rAF-driven node enter transition stalls (nodes stuck at `scale(0.85)`) and the `boundingBox`-based drag specs grab the wrong point.

## How (CATE_E2E only — prod/dev unchanged)

- **Never map a window to a display** — `revealWindow` is a no-op under e2e; Playwright drives the renderer over CDP. Plus `app.dock.hide()` and skipping the open-path `win.focus()`. `paintWhenInitiallyHidden` + `backgroundThrottling:false` keep the hidden renderer painting and firing `ready-to-show`.
- **Make the renderer deterministic without a visible window:**
  - a global stylesheet zeroes all `transition`/`animation` durations (installed by the e2e harness),
  - canvas nodes are created already `idle` (no rAF-driven enter animation, so geometry is final immediately),
  - node removal finalizes immediately instead of after the 200 ms exit delay.

`IS_E2E` gates every branch — prod and `npm run dev` take the exact same `show()`/`focus()` paths and animations as before.

## Verification

CI gates (all green locally):
- `npm run build` ✅
- `npm run typecheck` ✅
- `npm run test:unit` ✅ (472 passed, 3 skipped)
- `npm run test:smoke:electron` ✅

e2e drag/perf specs (not in CI; timing-sensitive, **already flaky on `origin/main`**): **4 failed across two runs = the `origin/main` baseline** — no regression — while the windows no longer appear on screen.


Anton-Horn:
Good PR. Very annoying indeed. Will merge once ci ran through.

architawr:
@Anton-Horn heads up before you merge — I force-pushed a **materially different approach** since your comment, and updated the title + description. Quick summary of what changed:

- The version you saw used `showInactive()`, but that still puts the window on screen. Testing showed there's **no macOS window-positioning trick that is both invisible and keeps the drag-spec timing**: off-screen coords get clamped back to ~`[0,34]`, and a hidden window throttles rAF so the node-enter animation stalls and `boundingBox`-based drag specs grab the wrong point.
- **New approach (CATE_E2E only):** the window is never shown at all (Playwright drives the renderer over CDP) + `app.dock.hide()`, and the renderer is made deterministic without a visible window — a global stylesheet zeroes all transitions/animations, canvas nodes are created already `idle` (no rAF-driven enter), and node removal finalizes immediately.
- **Result:** windows never appear on screen / never steal focus, and the (non-CI, already-flaky) drag/perf specs stay at the `origin/main` baseline (~4 flaky fails across two local runs). Prod/dev paths are unchanged — `IS_E2E` gates every branch.

`build` / `typecheck` / `test:unit` / `test:smoke:electron` all green, and the branch is re-synced with `main`. Full details in the updated description — worth a re-skim before merging since the diff is now broader than the original.

PaulHorn (approved):
## Review — quality pass ✅ (one non-blocking doc nit)

Reviewed the full diff (4 files, +62/-10). The change is well-scoped, correctly gated behind `IS_E2E` / `CATE_E2E`, and prod + `npm run dev` paths are untouched. CI build jobs are green on all three OSes.

### Correctness — sound
- **`revealWindow` no-op under e2e** is the right primitive; centralizing `show()`/`focus()` through it covers both the `ready-to-show` path and the new-window path (`{ focus: true }`).
- **`paintWhenInitiallyHidden: true` + `backgroundThrottling: false`** correctly let a never-shown window still fire `ready-to-show` and keep painting. The PR honestly documents that a hidden window *still* throttles rAF somewhat — which is exactly why the three renderer-side determinism fixes exist:
  - global CSS zeroing `transition/animation` durations (`e2eHarness`),
  - nodes created `idle` instead of `entering` (`canvasStore`) so geometry is final immediately,
  - exit finalized at `0ms` instead of `200ms` (`CanvasNode`).
  These compensate coherently for the throttled compositor. Logic checks out.
- Guards are correct: `canvasStore` uses `typeof window !== 'undefined' && window.electronAPI?.isE2E === true`; the drag test setup pins `isE2E: false`, so unit tests keep the `entering` path. `app.dock?.hide()` is safely optional-chained for non-macOS.

### Non-blocking nit — stale comment contradicts the implementation
The block comment above `IS_E2E` in `src/main/index.ts` describes an **abandoned** approach:

> "…under e2e we SHOW the window … but make it fully transparent and inactive — invisible to the user and never focused … Transparency (vs an off-screen position) sidesteps macOS clamping…"

But the actual code never shows the window:
```js
function revealWindow(win, opts = {}) {
  if (IS_E2E) return // never map to a display
  win.show(); if (opts.focus) win.focus()
}
```
The PR title/description confirm the final approach is **hidden** ("revealWindow is a no-op"), not transparent. The inline `webPreferences` comment ("the window is shown transparent + inactive") has the same drift. Functionally harmless, but misleading for future maintainers — worth a follow-up cleanup to match the comments to the "hidden" reality.

### Minor observation (out of scope)
`dragGhostWin.showInactive()` (drag-out ghost) is not gated by `IS_E2E`, so it would still map to a display during an OS-level panel detach. The targeted drag/perf specs drag canvas nodes (not OS drag-out), so this doesn't affect them — noting only for completeness.

**Verdict:** functionally correct, no regressions, CI green. Approving. The only cleanup is the stale "transparent" comments.

PR #251, opened 2026-06-02
feat(canvas): add Text element

verkhozin:
## Summary

Adds a lightweight **Text** element that lives directly on the canvas (Figma/Miro-style), for quick labels and notes alongside terminals, editors, and browsers.

- Editable text with per-element **text color**, **fill color**, **font size**, and **font family**, edited from a compact floating toolbar that appears above the element while it's active.
- Renders chrome-free (no tab bar / border / window background) so it reads as plain text on the canvas — distinct from the windowed panels.
- A small **×** button (top-right, on hover/active) deletes the element.

cate-feat-textfield

## Implementation

- New `text` `PanelType` wired through the existing two-touch panel registration: `PANEL_DEFINITIONS` (`src/shared/panels.ts`) + the renderer registry (`src/renderer/panels/registry.ts`), plus a `createText` factory and an `updateTextPanel` action in `appStore`.
- New `TextPanel` component (`src/renderer/panels/TextPanel.tsx`).
- A dedicated bare-rendering branch in `CanvasNode` for `text` nodes that skips the window chrome but reuses the existing move / resize / focus / z-order machinery.
- Creatable from the canvas toolbar "More" menu and the command palette ("New Text").
- Content, styling, and position persist to `.cate/workspace.json` (position via the standard canvas-node origin/size) and restore on load.

## Testing

- `npm run typecheck` — clean
- `npm run test:unit` — 532 passed
- `npm run build` — succeeds

## Notes

Happy to adjust naming, default sizes, the toolbar UX, or split this up if you'd prefer a different shape.

### DIFF
--- a/src/renderer/canvas/CanvasNode.tsx
+++ b/src/renderer/canvas/CanvasNode.tsx
@@ -509,6 +509,100 @@ const CanvasNode: React.FC<CanvasNodeProps> = ({
 
   if (!node) return null
 
+  // ---------------------------------------------------------------------------
+  // Bare text node — text elements live directly on the canvas with no window
+  // chrome (no tab bar, border, or background). The whole block is the drag
+  // handle when unfocused; when focused it becomes editable. Resize handles
+  // show only while active. Everything else (move, resize, focus, z-order,
+  // persistence) reuses the normal canvas-node machinery.
+  // ---------------------------------------------------------------------------
+  if (primaryPanelType === 'text') {
+    const textPanelId = primaryPanel?.id ?? node.panelId
+    const active = isFocused || isSelected
+    return (
+      <>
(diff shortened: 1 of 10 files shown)


PaulHorn:
Hey Egor, thanks a lot for putting this together. The work is clearly solid and the write-up is really thorough, which I appreciate.

That said, I'm going to pass on this one for now. A canvas-native text element is something we want to think through more carefully as part of the broader canvas direction, rather than land as a standalone piece, and I'd rather not have it sitting in the codebase half-aligned with where we're headed.

This isn't a reflection on the quality of the work at all. If we pick this up later I'll definitely circle back to it. Thanks again for taking the time, and sorry for the outcome here.

PR #287, opened 2026-06-05
feat(worktrees): spatial worktrees — territory layer, live sync, in-tab pill, session persistence

PaulHorn:
This branch grew well beyond its original icon-sharpness fix into the spatial-worktrees feature. Summary of everything it now contains.

## Spatial worktrees (canvas)
- Worktree **territory layer** (`territoryRenderer` / `territoryConfig` / `WorktreeTerritoryLayer`) replacing the old hull layer (`worktreeHull/` removed). Sludge field, per-worktree highlighting, and focus lens.
- `WorktreePill` / `Canvas` / `useCanvasNodeStyle` adjustments to drive the territory tint/highlight from the active tab's worktree.

## Live parallel-work sync
- New `worktreeSync` module + `useProcessMonitor` wiring keep the worktree registry current **in the background**, so parallel-work items refresh without opening the sidebar (and on app start). Piggybacks on the existing git monitor's fs-watch + adaptive poll — no new polling.
- `ParallelWorkTab.reconcile()` slimmed to share that sync and keep only the sidebar-only per-worktree status / PR fetches.

## Worktree pill in the active tab
- Moved the worktree chip out of the node toolbar's trailing controls into the **active tab** (`DockTabBar`), next to its title; removed from `CanvasNode` controls. Self-hides for non-terminal/agent panels and single-worktree workspaces.

## Session persistence (the part that was silently dropped before)
- Persist worktree **colors/labels** and each terminal/agent panel's **worktree tag** so they survive a restart — previously colors were re-rolled from the palette and panels came back tagged to the primary worktree even though they respawn inside their checkout.
- All of it lives in machine-local `session.json` (the checkouts under `.cate/worktrees` are gitignored), keeping the committed `workspace.json` free of runtime worktree ids:
  - `SessionSnapshot`/`ProjectSessionFile.worktrees` — the registry (absolute paths, like `workingDirectory`).
  - `ProjectSessionNode.worktreeId` — canvas-node terminal/agent tags.
  - `ProjectSessionFile.dockPanelWorktreeIds` — dock-zone terminal/agent tags (side map, keyed by panel id).
  - `appStore.hydrateWorktrees` merges by checkout path on restore so a saved color/label/id wins over anything a background sync already discovered.
- Tests: `worktreePersistence.test.ts` (restore hydration, merge-by-path precedence, save/restore round-trip incl. dock panels).

## Original icon-sharpness fix (kept)
Canvas world div `will-change: transform` made transient so vector icons re-rasterize crisply at rest at every zoom level. Single-file change in `Canvas.tsx`.

## Other
- `PI_VERSION` → 0.75.5.
- `mockups/` — design HTML for the worktree field/sludge/touched-files visualizations.

## Verification
- `tsc --noEmit` clean (pre-existing unrelated missing-module errors only).
- Worktree persistence + workspace-isolation + gitWorktree suites pass.


PaulHorn:
Pushed a follow-up commit: **GPU-render the territory backgrounds via WebGL2**, plus focus-lens dimming and a settings opt-out.

- **Performance:** territory now renders in a WebGL2 fragment shader — pan/zoom/drag are uniform updates + one quad draw at full res, no CPU tile-cache. The field is a pure function of world position, so view changes need no rebuild.
- **Fidelity:** the shader reproduces warp + smin SDF + softmax colour blend + analytic terraces/outlines + panel punch-out; visual constants are injected from `territoryConfig` so GPU and CPU can't drift. CPU `drawTerritory` is kept as a verified fallback (no WebGL2 / context lost) and A/B reference (`localStorage.cate.territory.backend = 'gl' | 'cpu'`).
- **Focus lens** now dims non-focused worktree territories (0.5 opacity + desaturate) to match the panels.
- **Opt-out:** Settings → Canvas → "Worktree territories" (default on); the layer fully unmounts when off.
- **Note:** enclosed-pocket fill is disabled on the GPU path — a non-local flood fill can't be reconciled with per-pixel rendering without blocky bleed; an enclosed lake reads as a clean clearing instead.

Added `territoryPocketMask` unit tests. `tsc`/build/tests green.

PR #366, opened 2026-06-09
feat(canvas): multi-node selection with marquee, group drag, and bulk close

### LINKED ISSUE #363: New terminal panel breaks existing terminal on click
### Problem

When a new terminal panel is created and placed anywhere around an existing terminal, clicking on the new terminal causes the old terminal to break (becomes unresponsive / loses its shell session).

### Steps to Reproduce

1. Open a terminal in Cate
2. Create a second terminal panel adjacent to it (split right / split down)
3. Click on the newly created terminal
4. Observe the original terminal

### Expected Behavior

Both terminals remain independently functional — clicking one should not affect the other.

### Actual Behavior

The original terminal breaks when the new terminal gains focus.

### Environment

- OS: macOS
- Cate version: 1.2.7

### LINKED ISSUE #364: Explorer Reload/Force reload has no effect for externally-modified files
### Problem

When an LLM edits a markdown file, the changes are correctly written to disk (confirmed by reading the file externally), but the Explorer markdown preview does not reflect the updated content. Neither "Reload" nor "Force reload" resolves the stale view. Even committing the file via git does not trigger a refresh in the explorer.

### Steps to Reproduce

1. Open a `.md` file in the Cate Explorer
2. Use opencode CLI to have an LLM edit that file
3. Verify the file was written correctly on disk (e.g., `cat` the file in a terminal)
4. Press "Reload" and/or "Force reload" in the Explorer
5. Observe the markdown preview

### Expected Behavior

The Explorer preview shows the latest file content after a reload — especially a forced reload.

### Actual Behavior

The Explorer shows stale content and does not pick up the on-disk changes even after explicit reload actions, including after a git commit.

### Environment

- OS: macOS
- Cate version: 1.2.7

### LINKED ISSUE #365: Canvas pan/zoom ergonomics regression from v1.1.1
### Problem

In version 1.1.1, panning was done with the select tool and zooming used a two-finger pinch gesture — both felt natural. The current version changed zoom to follow the cursor position, which is considerably more difficult to use for navigating around the canvas.

### Steps to Reproduce

1. Open a diagram or canvas with enough content to need panning and zooming
2. Try to pan around the canvas
3. Try to zoom in/out

### Expected Behavior

- Select tool should allow click-and-drag to pan
- Zoom should use two-finger pinch (trackpad) or be configurable between cursor-zoom and center-zoom

### Actual Behavior

- Panning is awkward without the select-tool drag behavior
- Zoom is cursor-position-based, making precise navigation difficult

### Environment

- OS: macOS
- Cate version: 1.2.7

Anton-Horn:
## What

Canvas multi-selection:
- Marquee drag over empty canvas selects multiple nodes
- Dragging any selected node moves the whole group together (`useGroupNodeDrag`)
- A selection ring marks selected-but-not-active panes (distinct from the focus halo)
- Canvas mini-dock context menus collapse to a bulk "Close All" while a multi-selection is active (`useMultiNodeSelection`)

## Also in here

- **terminal:** rebuild the shared WebGL glyph atlas across all live terminals on repaint, fixing scrambled glyphs in sibling terminals that share the atlas
- **editor:** reload the Monaco model when the file changes on disk externally (other editor, agent, git checkout), skipping while the buffer is dirty so unsaved edits are never clobbered
- **canvas:** select-tool trackpad two-finger scroll pans again (mouse wheel still zooms); freeze in-flight zoom/pan animations when anchoring a marquee so the box stays under the cursor
- node cleanup when a tab is moved out via "Move into New Window" leaves an empty node

Bumps version to 1.3.0.

Closes #363
Closes #364
Closes #365

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "cate",
-  "version": "1.2.7",
+  "version": "1.3.0",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "cate",
-      "version": "1.2.7",
+      "version": "1.3.0",
       "hasInstallScript": true,
       "license": "MIT",
       "dependencies": {
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "cate",
-  "version": "1.2.7",
+  "version": "1.3.0",
   "productName": "Cate",
   "description": "An infinite zoomable canvas IDE",
   "license": "MIT",
--- a/src/renderer/canvas/CanvasNode.tsx
+++ b/src/renderer/canvas/CanvasNode.tsx
@@ -18,6 +18,7 @@ import { useDragStore, useDragSourceVisibility } from '../drag'
 import { useNodeResize } from '../hooks/useNodeResize'
 import { useCanvasNodeStyle } from './useCanvasNodeStyle'
 import { useCanvasNodeDrag } from './useCanvasNodeDrag'
+import { useGroupNodeDrag } from './useGroupNodeDrag'
 import { useNodeResizeCursor } from './useNodeResizeCursor'
 import { NodeResizeOverlay } from './NodeResizeOverlay'
 import type { DockStore } from '../stores/dockStore'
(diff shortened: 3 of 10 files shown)


GHAFRI:
> editor: reload the Monaco model when the file changes on disk externally (other editor, agent, git checkout), skipping while the buffer is dirty so unsaved edits are never clobbered

This issue not fixed  in my use case, only solution I found to make it reload is to simply rename the file in the file explorer. Otherwise inside the canvas or opening the file in new window does not reload it. Force reload (Shift+CMD+R) does actually reload the md file but I would lose agent sessions which is not ideal

PR #367, opened 2026-06-09
fix: respect terminal font override settings

### LINKED ISSUE #285: Terminal "Font family override" setting is never applied (Nerd Font icons can't render)
## Summary

The **Settings → Terminal → "Font family override"** field (`terminalFontFamily`) is written to the settings store and persisted, but **never read** when constructing or updating the xterm.js terminal. The terminal font is hardcoded, so the override silently does nothing. A user who sets a Nerd Font (e.g. `FiraCode Nerd Font Mono`) to get powerline/icon glyphs sees no change — icons stay as tofu boxes.

## Environment

- Cate v1.1.1 (also present on current `main`, 846cb6f)
- macOS (darwin-arm64), Nerd Font installed system-wide
- Integrated terminal: `@xterm/xterm` + WebGL addon

## Root cause

`src/renderer/lib/terminal/terminalRegistry.ts` hardcodes the font in **both** xterm constructors and never consults the store:

- [`getOrCreate` constructor — L442](https://github.com/0-AI-UG/cate/blob/846cb6f35070cc230b00f5d056359e81ede759c0/src/renderer/lib/terminal/terminalRegistry.ts#L442)
  ```ts
  fontFamily: 'Menlo, Monaco, "Courier New", monospace',
  ```
- [`reconnectTerminal` constructor — L645](https://github.com/0-AI-UG/cate/blob/846cb6f35070cc230b00f5d056359e81ede759c0/src/renderer/lib/terminal/terminalRegistry.ts#L645)
  (same literal)

The live-settings subscription at [L367](https://github.com/0-AI-UG/cate/blob/846cb6f35070cc230b00f5d056359e81ede759c0/src/renderer/lib/terminal/terminalRegistry.ts#L367) applies `cursorBlink`, `scrollSensitivity`, `minimumContrastRatio` and `macOptionIsMeta` to live terminals, but has **no `fontFamily` branch** — so the override is discarded even on a fresh terminal.

A grep for `terminalFontFamily` returns only: the settings UI (write), the type declaration ([types.ts L1089](https://github.com/0-AI-UG/cate/blob/846cb6f35070cc230b00f5d056359e81ede759c0/src/shared/types.ts#L1089)), the default ([types.ts L1192](https://github.com/0-AI-UG/cate/blob/846cb6f35070cc230b00f5d056359e81ede759c0/src/shared/types.ts#L1192)), and the schema — **never a consumer**.

## Steps to reproduce

1. Install a Nerd Font (e.g. FiraCode Nerd Font Mono).
2. Settings → Terminal → set "Font family override" to `FiraCode Nerd Font Mono`.
3. Open a terminal; run a prompt with icons, or:
   ```
   echo -e "\ue0b0 \uf07c \ue702 \uf09b"
   ```

**Expected:** glyphs render in the Nerd Font.
**Actual:** font is unchanged (Menlo); Nerd Font PUA glyphs show as tofu boxes.

## Suggested fix

Read the setting (with the current literal as fallback) in both constructors, and add a `fontFamily` branch to the settings subscription so live terminals update — mirroring the existing `minimumContrastRatio` pattern. Sketch:

```ts
const DEFAULT_TERMINAL_FONT_FAMILY = 'Menlo, Monaco, "Courier New", monospace'
function getFontFamily(): string {
  return useSettingsStore.getState().terminalFontFamily?.trim() || DEFAULT_TERMINAL_FONT_FAMILY
}
// both constructors:  fontFamily: getFontFamily(),
// + applyFontFamilyToAll() wired into useSettingsStore.subscribe(...)
```

`reconnectTerminal`'s comment claims "same config as getOrCreate" — worth keeping the two genuinely in sync (the reconnect path is the cross-window transfer case).

HarunOr:
Closes #285

## Summary

- Apply `terminalFontFamily` when constructing xterm terminals.
- Apply `terminalFontSize` instead of hardcoding `13`.
- Keep terminal zoom render-scaling based on the configured base font size.
- Add regression coverage for default and configured terminal font settings.

### DIFF
--- a/src/renderer/lib/terminal/terminalLifecycle.ts
+++ b/src/renderer/lib/terminal/terminalLifecycle.ts
@@ -24,6 +24,8 @@ import {
   type RegistryEntry,
 } from './registryState'
 import {
+  getTerminalFontFamily,
+  getTerminalBaseFontSize,
   getScrollback,
   getScrollSensitivity,
   getContrastRatio,
@@ -103,8 +105,8 @@ export function createAndConfigureXtermTerminal(opts: CreateOpts): ConfiguredTer
 
   const terminal = new Terminal({
     theme: getActiveTheme().terminal,
-    fontFamily: 'Menlo, Monaco, "Courier New", monospace',
-    fontSize: 13,
+    fontFamily: getTerminalFontFamily(),
+    fontSize: getTerminalBaseFontSize(),
     cursorBlink: effectiveCursorBlink(),
     allowProposedApi: true,
     scrollback: getScrollback(),
--- a/src/renderer/lib/terminal/terminalRegistry.test.ts
+++ b/src/renderer/lib/terminal/terminalRegistry.test.ts
@@ -18,6 +18,16 @@ import { describe, it, expect, vi, beforeEach } from 'vitest'
 const events: string[] = []
 beforeEach(() => { events.length = 0 })
 
+const settingsState = {
+  terminalFontFamily: '',
+  terminalFontSize: 0,
+  terminalScrollback: 2000,
+  terminalCursorBlink: false,
+  terminalScrollSpeed: 1.0,
+  terminalContrast: 4.5,
+  terminalOptionIsMeta: true,
+}
+
 vi.mock('@xterm/xterm', () => {
   // Faithful-enough fake: models the buffer viewportY/baseY scroll indices and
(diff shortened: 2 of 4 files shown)


Anton-Horn:
Thanks for the PR. Well scoped, seems like the settings slipped through after some refactoring. Will merge once CI ran through.

PR #402, opened 2026-06-18
Add OSC52 terminal clipboard support

### LINKED ISSUE #403: Terminal should handle OSC52 clipboard writes
## Summary
Remote terminal apps such as tmux and Neovim can emit OSC 52 sequences to write copied text to the local clipboard. Cate currently renders the terminal through xterm.js but does not handle OSC 52 clipboard sequences, so remote copy operations are ignored.

## Repro
1. Open a Cate terminal.
2. SSH into a remote host.
3. Run tmux with clipboard support enabled, for example `set -g set-clipboard on`.
4. Copy text in tmux copy-mode or from an app that emits OSC 52.

## Expected
The copied text is written to the local system clipboard.

## Actual
The OSC 52 sequence is ignored and the local clipboard does not change.

## Notes
This affects remote tmux/neovim workflows where users expect terminal clipboard integration to work across SSH.

## Proposed fix
Register an xterm OSC 52 handler and route accepted clipboard writes through Electron main-process clipboard IPC.

Tracking PR: #402

seilk:
## Summary
- Register an xterm OSC 52 handler for terminal output.
- Decode clipboard-targeted OSC 52 payloads and write them through Electron main-process clipboard IPC.
- Add coverage for OSC 52 decoding/registration and the new clipboard IPC bridge.

Fixes #403

## Root cause
Cate used xterm.js for terminal rendering but did not handle OSC 52 sequences, so remote tmux/neovim clipboard writes were ignored. PTY-originated output cannot reliably use `navigator.clipboard` because browser clipboard writes may require user activation, so this patch routes accepted OSC 52 clipboard writes through Electron's main-process `clipboard.writeText`.

## Security note
OSC 52 lets terminal programs request system clipboard writes. This matches the behavior users expect when enabling tmux `set-clipboard`, but it means remote shells inside Cate can update the local clipboard.

## Validation
- `bunx vitest run src/renderer/lib/terminal/terminalOsc52Clipboard.test.ts src/main/ipc/terminal.test.ts src/main/ipc/ipcConformance.test.ts`
- `bun run typecheck`
- `bun run lint` (0 errors, existing warnings only)
- `bun run build`
- `bun run test` (181 passed, 1 skipped)

### DIFF
--- a/src/main/ipc/terminal.test.ts
+++ b/src/main/ipc/terminal.test.ts
@@ -22,7 +22,11 @@ vi.mock('node-pty', () => ({
 
 // Captured ipcMain.handle map so tests can invoke a handler directly.
 const handlers = new Map<string, (...args: unknown[]) => unknown>()
+const clipboardWriteText = vi.fn()
 vi.mock('electron', () => ({
+  clipboard: {
+    writeText: clipboardWriteText,
+  },
   ipcMain: {
     handle: vi.fn((channel: string, fn: (...args: unknown[]) => unknown) => {
       handlers.set(channel, fn)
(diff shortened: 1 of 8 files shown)


Anton-Horn:
write-only OSC 52, ignores read/paste so no clipboard exfil. clean and tested, CI green. merging.

PR #416, opened 2026-06-21
Stateful cate.agent conversation API

Anton-Horn:
Stacked on #397.

Replaces the one-shot `cate.agent.run` with a turn-based session API so an extension can hold a real multi-turn conversation:

- `open({ resume? })` returns a handle (pi's session file path)
- `send(sessionId, prompt)` runs one turn, returns the full assistant message `{ text, message }`
- `dispose(sessionId)` tears down the live client; pi's jsonl stays, so the handle resumes later
- `run(prompt)` stays as open -> send -> dispose sugar

pi owns all conversation state on its session file; Cate only holds the in-memory client, exactly like a panel (same `create()`/`PiRpcClient`/`dispose()` path). One live session per extension, one turn in flight: the anti-runaway guard.

Also fixes the bug behind every "(no text)": a turn that ends on `stopReason: 'error'` (unsupported model, auth, bad request) was indistinguishable from an empty turn, so it returned silent empty text. It now rejects with pi's reason and the handler surfaces it to the extension as `failed: <reason>`.

Kitchen Sink demo that exercises this: 0-AI-UG/cate-extensions#6.

Tests: open/send/dispose dispatch + consent gating, turn runner (willRetry skip, answer extraction, error-turn rejection). All green.

### DIFF
--- a/src/agent/main/agentManager.test.ts
+++ b/src/agent/main/agentManager.test.ts
@@ -78,3 +78,79 @@ describe('AgentManager.disposeForWebContents', () => {
     expect(disposed.sort()).toEqual(['b', 'c'])
   })
 })
+
+// runTurn is the non-streaming extension turn runner: it resolves with the final
+// assistant message read straight off the terminal `agent_end` event's `messages`.
+// The tricky case is pi's auto-retry: it emits an `agent_end` with willRetry:true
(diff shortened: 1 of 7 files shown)


Anton-Horn:
Auto-marked merged: the commit was fast-forwarded into feat/extension-system, so it ships as part of #397. No separate review needed here.

PR #420, opened 2026-06-22
fix(main): reveal window when renderer load completes

FingerCaster:
## Summary
- reveal the BrowserWindow once either `ready-to-show` or `did-finish-load` fires
- run main-window startup readiness/deferred init through the same once-only fallback
- apply the same readiness fallback when macOS `activate` recreates the main window

## Testing
- npm run typecheck
- npm test -- src/main/lifecycle/shutdown.test.ts
- npm run build
- npm run test:smoke:electron

### DIFF
--- a/src/main/index.ts
+++ b/src/main/index.ts
@@ -355,14 +355,19 @@ app.whenReady().then(async () => {
   }
 
   // Check for a crash report from the previous session — shows an opt-in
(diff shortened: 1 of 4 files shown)


Anton-Horn:
added a regression test for the did-finish-load fallback (fails without the fix, passes with it). CI green, merged. thanks!

PR #422, opened 2026-06-26
feat(sidebar): show installed skills as chips in the Workspaces view

FlaminJoe:
## What

Adds a compact **SkillsQuickBar** at the bottom of the Workspaces view (just above the version marker) that lists the skills installed in the current workspace as small chips.

- **Hover** a chip → tooltip names the agents it's installed for (e.g. `grill-me · Claude Code, Cate Agent`).
- **Click** a chip or the **+** → opens the existing `SkillsDialog` (browse / install / save).
- **Empty state** doubles as discovery: `No skills yet — add some`, which also opens the dialog.

## Why

Right now skills are only visible *inside* `SkillsDialog`, opened from the puzzle button in the activity bar. Once a workspace has skills there's no at-a-glance reminder of what its agents can already do. This surfaces them where you already look — next to the project list.

## How

- New `SkillsQuickBar.tsx`, rendered from `Sidebar.tsx` under the same `isExpanded && activeView === 'workspaces'` guard as the version marker, so it follows the Workspaces view to whichever side hosts it.
- Reads the workspace manifest via the **existing** `skillsListInstalled` IPC — no new channels, no main-process changes. It refetches when the Skills dialog closes so installs/uninstalls there reflect immediately.
- **Read-only**: it never installs, writes, or mutates anything.
- The pure folding logic (one row per skill×target → one chip per skill, agents deduped, sorted by name) lives in `skillChips.ts` with unit tests, matching the existing `*.ts` + `*.test.ts` pattern in `src/renderer/sidebar/`.
- Styling reuses existing tokens/classes (`text-muted`, `border-subtle`, `bg-white/5`, `Tooltip`) to match the sidebar chrome. Chip list is capped (`max-h-24 overflow-y-auto`) so it never crowds the project list.

## Verification

- `tsc --noEmit` — clean
- `eslint` on changed files — clean
- `vitest run skillChips.test.ts` — 5/5 passing

Built against `main` @ `538db77` (1.3.2). I couldn't run the full Electron app in this environment, so a maintainer sanity-check of the live render is welcome.


Anton-Horn:
Pushed a reshape: skills now live in a collapsible Skills node inside each workspace (grouped by agent), instead of the bottom bar. Also made canvases collapsible, fixed the Cate agent icon, and added a footer link to sources.json for adding missing skills.

PR #479, opened 2026-07-18
Restore agent sessions across app restarts

Anton-Horn:
Terminals running an agent CLI (claude, codex, pi, opencode, cursor-agent) now remember which session the agent had open, and session restore relaunches the agent with the right resume flag instead of dropping to a plain shell.

The probe lives in the runtime daemon so local and remote workspaces behave the same. Per CLI it uses the strongest available signal: claude's pid registry, codex's open rollout fd, and start-time-gated store scans for the rest.

Store shapes are pinned by an opt-in live contract suite (`npm run test:agent-contracts`) that drives the real installed CLIs, plus a weekly launchd schedule to catch CLI updates early.


Anton-Horn:
Swapped the live contract suite: instead of reverse-engineering session stores it now pins each CLI's hook surface (session id, turn status, env echo-back) plus resume-by-id. All 9 tests green in ~1 min against the real CLIs. Store-join steps keep the fallback probe's parsers covered.

PR #518, opened 2026-07-22
fix(security): gate repo-controlled layout restore behind workspace trust

Anton-Horn:
Fixes GHSA-8769-jp52-985f (reported by @glmgbj233).

A committed `.cate/workspace.json` could restore an agent panel on open, which starts pi, which runs the repo's eager `.pi/mcp.json` command. Code execution from just opening a repo.

Untrusted projects now restore passive layout only (canvas, dock, editors inside the project). Banner offers the opt-in. Trust lives in userData so a repo can't vouch for itself. Also gates the MCP adapter on the same check and pins it to 2.11.0.

e2e opens the real PoC repo in a real app. I reverted the gate to confirm the tests actually fail without the fix.


Anton-Horn:
Simplified the model: trust and open, or don't open. No more restricted mode, so the layout filter and the withheld notices are gone (-640 lines). The gate now runs before we read the project's .cate/ files.

PR #565, opened 2026-08-05
fix: handle mapped network drives in path validation (#239)

### LINKED ISSUE #239: Workspaces on mapped network drives never load — file explorer stays empty
# Workspaces on mapped network drives never load — file explorer stays empty ("outside allowed directories")

## Summary

Opening a workspace on **any mapped network drive** (regular SMB shares, rclone/WinFsp mounts, sshfs-win) results in a permanently empty file explorer. Depending on the mount type, the workspace is either rejected outright or accepted but unusable. The root causes are in the path-validation layer (`validatePath` / `isWithinAllowedRoots` / `resolveTrustedWorkspaceRoot`).

This also blocks #216 (WSL workspaces) — `\\wsl$\...` paths are UNC paths and will hit the same bugs.

I debugged this down to four distinct root causes, with code references and suggested fixes below.

## Environment

- Cate v1.1.0 (Electron 41.7.1, Node 24.15.0)
- Windows 11 Pro, Build 26200
- Reproduced with: regular SMB mapped drives, rclone mount (WinFsp), sshfs-win

## Reproduction

**Simplest repro (any Windows machine with a file server):**

1. Map a network drive: `net use X: \\fileserver\share`
2. In Cate: *Open Folder* → pick any folder on `X:`
3. → Explorer shows "No files found" forever; `main.log` fills with `Access denied: path "X:\..." is outside allowed directories`

**Repro without a file server (rclone, free):**

1. `winget install Rclone.Rclone WinFsp.WinFsp`
2. `rclone config create demo sftp host=<any-ssh-host> user=<user> key_file=<key>`
3. `rclone mount demo:/home/<user> X: --network-mode --vfs-cache-mode writes`
4. Open a folder on `X:` in Cate → same result

## Root cause analysis

All code references are from `dist/main/index.js` in the v1.1.0 app.asar.

### Bug 1 — Native `fs.promises.realpath` fails on WinFsp disk-mode mounts → workspace rejected

`resolveTrustedWorkspaceRoot()` does:

```js
const resolved = path.resolve(rootPath);
const realPath = await fs.realpath(resolved);   // ← throws on WinFsp disk-mode drives
const stat = await fs.stat(realPath);
```

`fs.promises.realpath()` uses the **native** realpath (`GetFinalPathNameByHandle`). On WinFsp drives mounted in disk mode (the default for sshfs-win and `rclone mount` without `--network-mode`), this API fails because the volume is not registered with the Windows Mount Manager:

```
fs.promises.realpath("X:\")                 → FAIL: UNKNOWN: unknown error, realpath 'X:\'
fs.realpathSync("X:\")  (JS implementation) → OK:   X:\
```

Result: `catch` → `"workspaceRoots: rootPath does not exist or is unreadable, rejecting"` → workspace can never be opened.

**Suggested fix:** fall back when the native call fails:

```js
let realPath;
try {
  realPath = await fs.realpath(resolved);
} catch {
  realPath = fsSync.realpathSync(resolved);   // JS implementation, or just use `resolved`
}
```

### Bug 2 — Path-form mismatch: main registers UNC form, renderer requests drive-letter form (affects ALL mapped network drives, including plain SMB)

On any mapped network drive, `fs.realpath("X:\folder")` returns the **UNC form**:

```
fs.promises.realpath("Y:\")  →  \\fileserver\share\        (regular SMB share)
fs.promises.realpath("X:\")  →  \\server\volname\          (rclone --network-mode)
```

`createWorkspace()` stores **and registers** this UNC form:

```js
trustedRoot = await resolveTrustedWorkspaceRoot(rootPath);  // returns UNC form
info.rootPath = trustedRoot;
addAllowedRoot(info.rootPath);                              // allowedRoots = { "\\server\share\folder" }
```

…but the **renderer/file-explorer keeps requesting the drive-letter form** the user picked in the dialog:

```
[info]  Workspace created: 0cf5ab8d (\\server\volname\folder)        ← main: UNC registered
[error] [fs:readDir] Error: Access denied: path "X:\folder" is outside allowed directories   ← renderer: drive form
```

`validatePath()` compares with `path.resolve(filePath)` (no realpath), so `"X:\folder"` never matches the allowed root `"\\server\volname\folder"`. **Permanent mismatch — restarting does not help** (the persisted root is the UNC form, the explorer still uses the drive form).

Note: PR #212 (retry initial load) does not help here because every retry uses the same mismatched path form.

**Suggested fixes (either):**
- Register **both** forms as allowed roots (the resolved input path *and* its realpath), or
- Canonicalize consistently: make `validatePath` compare realpath-canonicalized forms on both sides (with a cache, since realpath per IPC call is expensive), or
- Don't canonicalize the workspace root to UNC at all — validate it, but store/register the user-picked resolved form.

### Bug 3 — Trailing-backslash bug in `isWithinAllowedRoots` for filesystem roots

```js
for (const root of allowedRoots) {
  if (normalized.startsWith(root + path.sep) || normalized === root) return true;
}
```

When the workspace root is a **filesystem root** (`X:\` or `\\server\share\`), `path.resolve()` keeps the trailing backslash. Then `root + path.sep` produces a **double backslash** (`X:\\`), which no child path starts with:

```js
root          = "X:\"               // path.resolve("X:\") keeps trailing sep
root + sep    = "X:\\"              // ← no real path starts with this
"X:\folder".startsWith("X:\\")      // false → child rejected
```

So even when Bugs 1+2 don't apply, opening a *drive root* as workspace shows the root but rejects every child ("No files found"). Verified by simulating the exact `validatePath`/`isWithinAllowedRoots` code against a live mount: workspace root passes, every child path fails.

**Suggested fix:**

```js
const rootWithSep = root.endsWith(path.sep) ? root : root + path.sep;
if (normalized === root || normalized.startsWith(rootWithSep)) return true;
```

### Bug 4 — Unbounded retry storm with unhandled rejections when `.cate` creation fails

When the workspace root is not writable (e.g. a read-only share, or a FUSE mount that maps `mkdir`-on-existing to `EPERM` instead of `EEXIST` — sshfs does this due to SFTPv3 limitations), the project-state save loop retries forever, ~1-2×/second, each time logging an unhandled rejection:

```
[error] Unhandled rejection Error: EPERM: operation not permitted, mkdir 'X:\.cate'
[warn]  [session] Project state save failed for X:\: ...
(repeats indefinitely, hundreds of entries per minute)
```

**Suggested fix:** treat persistent state-save failures as non-fatal (disable after N failures or exponential backoff), and handle `EPERM` from `mkdir` the same as `EEXIST` when the directory already exists.

## Impact

- Any user with a mapped corporate network drive cannot open workspaces on it (Bug 2 alone is enough).
- All WinFsp-based mounts (rclone, sshfs-win) are unusable as workspaces (Bugs 1+2).
- WSL workspaces (#216) will require the same fixes, since `\\wsl$\...` are UNC paths.

## Evidence / methodology

Found by debugging against a live SFTP mount: extracted the validation functions from `app.asar`, replayed them step-by-step with `ELECTRON_RUN_AS_NODE` against the mount, and correlated with `main.log`. Happy to provide full logs or test scripts if useful.

PiedPiper911:
## Problem
Workspaces on mapped network drives (SMB, rclone, WinFsp, sshfs-win) never load because path validation fails due to 4 distinct bugs:
1. `realpath` throws on FUSE/WinFsp paths
2. UNC paths don't match their mapped drive letter equivalents
3. Trailing backslash on filesystem roots causes validation failure
4. Read-only mounts trigger retry storms

## Solution
1. Wrap `realpath` in try/catch with fallback to original path
2. Normalize paths to a canonical form before comparison (resolve both UNC and drive-letter paths)
3. Handle filesystem root paths (trailing separator) correctly
4. Add retry limits for read-only mount detection

Fixes #239

### DIFF
--- a/src/main/ipc/pathValidation.test.ts
+++ b/src/main/ipc/pathValidation.test.ts
@@ -1,4 +1,5 @@
 import fs from 'node:fs/promises'
+import fsSync from 'node:fs'
 import path from 'node:path'
 import { afterEach, beforeEach, describe, expect, test, vi } from 'vitest'
 import {
@@ -37,6 +38,7 @@ describe('pathValidation', () => {
     await fs.rm(rootDir, { recursive: true, force: true })
     await fs.rm(outsideDir, { recursive: true, force: true })
     vi.useRealTimers()
+    vi.restoreAllMocks()
   })
 
   test('allows creation inside trusted roots', async () => {
@@ -132,6 +134,32 @@ describe('pathValidation', () => {
       await fs.symlink(outsideDir, link)
       await expect(validatePathStrict(path.join(link, 'new-file.txt'), undefined, SCOPE)).rejects.toThrow(
         /outside allowed directories/,
+      )
+    })
+  })
+
+  describe('realpath fallback', () => {
+    test('uses the JS resolver when the native resolver fails on an existing mount path', async () => {
+      const target = path.join(rootDir, 'file.txt')
+      await fs.writeFile(target, 'data')
+      vi.spyOn(fs, 'realpath').mockRejectedValueOnce(
+        Object.assign(new Error('unknown realpath error'), { code: 'UNKNOWN' }),
+      )
+
+      await expect(validatePathStrict(target, undefined, SCOPE)).resolves.toBe(
+        fsSync.realpathSync(target),
+      )
+    })
+
+    test('fails closed when both realpath implementations reject a non-ENOENT error', async () => {
+      const target = path.join(rootDir, 'file.txt')
+      await fs.writeFile(target, 'data')
+      const error = Object.assign(new Error('I/O failure'), { code: 'EIO' })
+      vi.spyOn(fs, 'realpath').mockRejectedValue(error)
+      vi.spyOn(fsSync, 'realpathSync').mockImplementation(() => { throw error })
+
+      await expect(validatePathStrict(target, undefined, SCOPE)).rejects.toThrow(
+        /cannot resolve real path/,
       )
     })
   })
--- a/src/main/ipc/pathValidation.ts
+++ b/src/main/ipc/pathValidation.ts
@@ -1,5 +1,5 @@
 // =============================================================================
-// Path validation — prevent path traversal and restrict filesystem access
+// Path validation  prevent path traversal and restrict filesystem access
 // to registered workspace roots and the system temp directory.
 // =============================================================================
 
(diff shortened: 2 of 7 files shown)


Anton-Horn:
Thanks for working on this. I think it needs another pass before merging: the non-`ENOENT` fallback makes strict validation fail open when `realpath` errors, potentially allowing symlink escapes. The original issue also remains partly unresolved—WinFsp mounts are rejected earlier in `workspaceRoots.ts`, drive-letter and UNC forms aren’t actually mapped, and the new traversal cap doesn’t limit `.cate` save retries. Could you add targeted tests for these cases?

Anton-Horn:
I pushed the remaining follow-up fixes and tests. This now safely handles the WinFsp fallback, drive/UNC aliases, and bounded `.cate` save retries. Looks good to merge.

PR #592, opened 2026-08-23
feat(dock): add Copy Path / Copy Relative Path to file-backed tab context menus

### LINKED ISSUE #591: Tab context menu has no Copy Path / Copy Relative Path
The file explorer's context menu has Copy Path / Copy Relative Path, but the tab context menu (dock tabs and canvas file windows) only offers Rename / Close / Split Right / Move into New Window. So copying the path of a file that's already open means going back to the explorer, finding its node, and right-clicking there.

Where this bites: a file created or edited on the canvas (a note, a log dump) whose path you want to paste into an agent prompt. The tab in front of you already knows its path; the detour through the explorer is the only reason to open the explorer at all in that flow.

**Request:** show Copy Path / Copy Relative Path in the tab context menu for file-backed panels, with the explorer's copy behavior.

The implementation turned out to be small — the tab menu hook can reuse the explorer's `parseLocator` / `relativeDisplayPath` as-is — so I'm opening a PR alongside this issue. Happy to adjust the menu placement or drop it if this isn't a direction you want.

wyddy7:
Closes #591

Adds **Copy Path** and **Copy Relative Path** to the tab context menu (dock tabs and canvas file windows), shown only for panels backed by a file.

Today these actions exist only in the file explorer's context menu. When a file is already open — especially as a canvas window — copying its path means going back to the explorer, finding the node, and right-clicking there. The tab you're looking at already knows its path; this removes the detour. The concrete workflow it serves: create/edit a note or log on the canvas, copy its path from the tab, paste it into an agent prompt.

- `Copy Path` — absolute host path (remote `cate-runtime://` locators copy the host path, not the internal URI, via the existing `parseLocator`)
- `Copy Relative Path` — workspace-relative via the existing `relativeDisplayPath`, falling back to the absolute path when the file sits outside the workspace root or no root is set

Implementation is a single gated group in `handleTabContextMenu` (`src/renderer/docking/useDockTabActions.ts`), so it covers side/center dock zones, canvas mini-docks, and detached dock windows in one place. Both helpers are reused as-is from the explorer path — no new path logic. The explorer's `Alt+Cmd+C` accelerators are display-only labels (bound nowhere), so they're intentionally omitted here rather than advertising chords that don't fire.

Purely additive: panels without a `filePath` (terminals, browsers, scratch editors) see the menu exactly as before.

Tests: `useDockTabActions.copyPath.test.tsx` (mirrors the existing rename-test harness) — items offered and absolute path copied; workspace-relative copy; remote-locator host path; no-rootPath fallback; items absent for a scratch editor.

Local checks (this branch): `npm run typecheck` clean; `npm run lint` clean on changed files; `npm test` = 3018 passed, 50 skipped — the 7 failures in the two sidebar tree-collapse suites reproduce identically on unmodified `main` (pre-existing, unrelated); `npm run build` succeeds.

Screenshot (canvas file window):

Tab context menu with Copy Path and Copy Relative Path

### DIFF
--- a/src/renderer/docking/useDockTabActions.copyPath.test.tsx
+++ b/src/renderer/docking/useDockTabActions.copyPath.test.tsx
@@ -1,0 +1,165 @@
+// =============================================================================
+// useDockTabActions tab menu — Copy Path / Copy Relative Path. The pair mirrors
+// the file explorer's items and is gated on panel.filePath: a scratch editor
+// has no file yet, so offering the items there would copy nothing. Relative
+// resolves against the workspace rootPath, falling back to the absolute host
+// path; remote locators copy the host path, not the cate-runtime:// URI.
+// =============================================================================
+
+import React from 'react'
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
+import { createRoot, type Root } from 'react-dom/client'
+import { act } from 'react'
+
+;(globalThis as unknown as { IS_REACT_ACT_ENVIRONMENT: boolean }).IS_REACT_ACT_ENVIRONMENT = true
+
+vi.mock('../lib/logger', () => ({ default: { info: vi.fn(), warn: vi.fn(), error: vi.fn(), debug: vi.fn() } }))
+vi.mock('../lib/terminal/terminalRegistry', () => ({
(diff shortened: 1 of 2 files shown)


Anton-Horn:
Thanks for the clean, focused change and the thorough tests — merged!

